PR #18, opened 2015-01-14
Scrapinghub API URL fixed

bosnj:
Just an update with the correct Scrapinghub API URL, to warm up the account :smile:


pablohoffman:
keep'em coming @bosnj ! :)

PR #22, opened 2015-12-04
updating README to fit with the API

stummjr:
I used the API and noticed that it behaves slightly different from the documentation. So, this PR is meant to update the documentation.

**Summary of changes**:
 - `project_ids()` returns a list of integers instead of a list of strings
 - `project.spiders()` returns a list of dicts with spider metadata instead of a list with the spider names
 - `schedule` and `project.id` return an id like `123/2/12` rather than an hexa string

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,25 +27,28 @@ First, you connect to Scrapinghub::
 You can list the projects available to your account::
 
     >>> conn.project_ids()
-    [u'123', u'456']
+    [123, 456]
 
 And select a particular project to work with::
 
-    >>> project = conn['123']
+    >>> project = conn[123]
     >>> project
-    Project(Connection('APIKEY'), '123')
+    Project(Connection('APIKEY'), 123)
     >>> project.id
-    '123'
+    123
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -27,25 +27,28 @@ First, you connect to Scrapinghub::
 You can list the projects available to your account::
 
     >>> conn.project_ids()
-    [u'123', u'456']
+    [123, 456]
 
 And select a particular project to work with::
 
-    >>> project = conn['123']
+    >>> project = conn[123]
```
eliasdorneles: I think both versions work. :)

stummjr: @eliasdorneles they do work, but since the `conn.project_ids()` returns a list of int, it looks more consistent if we use `conn[123]` rather than `conn['123']` in the doc.

eliasdorneles: fair enough!
LGTM, merging!

nramirezuy: I would have left an example using string since the most common task is passing the project id from the arguments of a script.

PR #23, opened 2016-03-30
merge python-hubstorage into python-scrapinghub

bertinatto:
Any comments?


vshlapakov:
Reworked with keeping the whole history for the both packages in https://github.com/scrapinghub/python-scrapinghub/pull/28.

PR #25, opened 2016-06-02
Add the 'run' endpoint

jesuslosada:
Deprecate the 'schedule' endpoint in favor of the new 'run' endpoint.


vshlapakov:
Thanks @jesuslosada! That's a shame that the PR got stuck in review for so long time, and it's late to change `scrapinghub.py` (as it's deprecated and was moved to `scrapinghub/legacy.py`), but I'll update it in the new client 👍

PR #27, opened 2016-09-29
Merge python-hubstorage as a subtree

vshlapakov:
1st step to merge the libraries with keeping history of both.
(Doesn't provide API backward compatibility, so far it's only about proper merging.)

The only drawback I spotted is that `git log` doesn't work as-is for the subtree, it's discussed and explained with details, with describing a workaround [here](http://stackoverflow.com/questions/10918244/git-subtree-without-squash-view-log). All the Git GUI clients show it properly, but Github itself doesn't unfortunately (`blame` works, `history` reflects only last merge commit), not sure if it's a problem for us.

Review, please.


vshlapakov:
Doesn't play nice with Github, closing in favour of https://github.com/scrapinghub/python-scrapinghub/pull/28.

PR #28, opened 2016-09-30
Merge python-hubstorage

vshlapakov:
1st step to merge the libraries with keeping history of both.

Rework of #27, based on [this](https://saintgimp.org/2013/01/22/merging-two-git-repositories-into-one-repository-without-losing-file-history/), looks like the current approach hasn't drawbacks of the subtree merge.
What I've done here is basically:
- moved scrapinghub repo files to a subfolder
- added hubstorage repo as remote and merged it to this branch
- moved hubstorage repo files to a subfolder
- moved scrapinghub repo files back to root directory
- sorted out common files, fixed scrapinghub tests

The PR provides API compatibility:
```
$ PYTHONPATH=./scrapinghub-1.8.0-py2.7.egg python
Python 2.7.12 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:42:40) 
>>> from hubstorage import HubstorageClient
>>> from scrapinghub import Job
>>>

$ PYTHONPATH=./scrapinghub-1.8.0-py3.5.egg python
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:53:06) 
>>> from scrapinghub import Job
>>> from hubstorage import HubstorageClient
>>>
```

Some important notes about reviewing the PR:
- these folders were moved as-is 

1. (origin py-hs repo)/hubstorage -> hubstorage
2. (origin py-hs repo)/tests -> tests/hubstorage
3. scrapinghub/tests -> tests/scrapinghub

- these files were moved from hubstorage repo root to scrapinghub repo root as-is:

1. LICENSE
2. Makefile
3. MANIFEST.in
4. README.rst -> README-hubstorage.rst
5. requirements-pypy.txt

- these files were merged/updated manually:

1. setup.py
2. requirements.txt
3. requirements-test.txt
4. tox.ini

I haven't deleted a branch with subtree merge attempt, so you can compare results:
https://github.com/scrapinghub/python-scrapinghub/tree/sc411/hubstorage

Review, please.


setup.py
```diff
@@ -14,8 +19,9 @@
     author_email='info@scrapinghub.com',
     url='http://github.com/scrapinghub/python-scrapinghub',
     platforms = ['Any'],
-    py_modules = ['scrapinghub'],
-    install_requires = ['requests'],
+    py_modules = ['scrapinghub', 'hubstorage'],
```
vshlapakov: Updated structure & fixed imports: now it fits the plan. Last thing is proper versioning - I guess we're going to have a single version for the merged repository. 
Should we do it in the same PR or update it later? I'd use Scrapy approach: 
- `VERSION` file in `scrapinghub/`
- `setup.py` / `scrapinghub.__init__` reading from `VERSION`

chekunkov: > I guess we're going to have a single version for the merged repository.
@vshlapakov sure. However, I'd like to postpone release a bit - e.g. to try version from master after merge in some internal services before publishing it to PyPI. So let's setup it properly but not change the version yet.

Regards approach - you can use approach from scrapy, just don't forget to include VERSION file in package data (or MANIFEST.in). Or you can stick with current setup with bumpversion - add __version__ variable to scrapinghub/__init__.py, mention this file and setup.py in bumpversion config.

We are going to bump python-scrapinghub version 1.8.0 -> 1.9.0 - right?

chekunkov: > from scrapinghub import ... should work as before (I think it's only APIError and Connection

btw, ^ my fault, for backwards compatibility we need everything from `legacy.py` to be importable from `scrapinghub`. So you need to move `__all__ = ["APIError", "Connection", "HubstorageClient"]` from `scrapinghub/legacy.py` to `scrapinghub/__init__.py`, and add `from .legacy import *` to `scrapinghub/__init__.py`. If someone has `from scrapinghub import Project` their code will be broken by current change.

chekunkov:
Another note - we need to merge READMEs, but I think that it might be easier to review and manage as a separate PR.

setup.py
vshlapakov: Done with versioning: bumpversion looks cleaner than reading from `VERSION` everywhere. Right, let's update version in master later after proper testing, +1. Also fixed imports, feel free to check.

scrapinghub/legacy.py
```diff
@@ -24,8 +26,6 @@
 
 
 __all__ = ["APIError", "Connection"]
```
chekunkov: you need to remove this line as well for `from .legacy import *` to work as we expect

scrapinghub/legacy.py
```diff
@@ -11,6 +11,8 @@
 import time
 import warnings
 
+from . import __version__
```
chekunkov: hm, doesn't that lead to a circular import? you import from `legacy` in `__init__` and import from `__init__` in `legacy`

vshlapakov: Oh, fixed, thanks

scrapinghub/__init__.py
```diff
@@ -0,0 +1,10 @@
+__all__ = ['__version__', "APIError", "Connection", "HubstorageClient"]
```
chekunkov: don't think `__version__` should be included here

scrapinghub/hubstorage/client.py
```diff
@@ -0,0 +1,182 @@
...
+def _get_package_version():
+    """Small helper to avoid circular imports"""
+    from scrapinghub import __version__
+    return __version__
+
+
+class HubstorageClient(object):
+
+    DEFAULT_ENDPOINT = 'https://storage.scrapinghub.com/'
+    DEFAULT_USER_AGENT = 'python-hubstorage/{0}'.format(_get_package_version())
```
chekunkov: default user-agent should be `python-scrapinghub/{version}`

scrapinghub/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+
+
+import pkgutil
+__version__ = pkgutil.get_data(__package__, 'VERSION').decode('ascii').strip()
```
chekunkov: btw, `__version__` was a string and should remain a string in both  py2 and py3. now it's a unicode in py2.

PR #29, opened 2016-10-11
Merge README of hubstorage and scrapinghub

vshlapakov:
Review, please (merge only after #28).


README.rst
```diff
@@ -5,18 +5,23 @@ Client interface for Scrapinghub API
...
+
+
+.. contents:: :depth: 1
+
+
 Requirements
 ============
 
 * Python 2.6 or above
 * `Requests`_ library
```
chekunkov: as well as 'retrying>=1.3.3', 'six>=1.10.0'. we either need to keep this section up to date with setup.py or remove it (e.g. I don't see such section in https://github.com/kennethreitz/requests or )

what is missing is Installation section with `pip install python-scrapinghub`, + mention that msgpack support can be enabled by installing `pip install python-scrapinghub[msgpack]`.

vshlapakov: You're right, though I'd leave Python itself in requirements and that's all.
Added Installation section 👍

PR #31, opened 2016-10-18
Fix ResourceType._allows_mpack check.

chekunkov:
Existing check is broken. It works for items, logs and samples only if one uses them through Job object. However it doesn't work correctly for `list(proj.items.apiget('33/5/stats'))`). Also for collections it applies the same rules as for items, but only few collections endpoints support msgpack - scan, batch scan and get collection item. Other endpoints do no support msgpack.

### DIFF
--- a/scrapinghub/hubstorage/collectionsrt.py
+++ b/scrapinghub/hubstorage/collectionsrt.py
@@ -1,12 +1,44 @@
 import re
+
 from requests.exceptions import HTTPError
+
 from .resourcetype import DownloadableResource
+from .serialization import MSGPACK_AVAILABLE
 from .utils import urlpathjoin
+
+
+COLLECTIONS_MSGPACK_REGEX = re.compile(
+    r"""(v?c?s)  # collection type
+        /\w+     # collection name
+        (
+            /?                 # no key
+            |                  # OR
+            /(?P<key>[^/]+)/?  # item key
+        )
+        $
+    """,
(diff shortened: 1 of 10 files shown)


scrapinghub/hubstorage/collectionsrt.py
```diff
@@ -1,13 +1,51 @@
...
+
+        :param path: None, tuple or string
+
+        """
+        if not MSGPACK_AVAILABLE:
+            return False
+        # path
+        path = urlpathjoin(path or '')
+        match = COLLECTIONS_MSGPACK_REGEX.match(path)
+        if match:
```
vshlapakov: `return (match and match.group('key') != 'count)'`?

vshlapakov:
Except minor proposal, LGTM 👍

chekunkov:
@vshlapakov does it look better now?

scrapinghub/hubstorage/collectionsrt.py
```diff
@@ -1,13 +1,45 @@
...
+        """Check if request can be served with msgpack data.
+
+        Collection scan and get requests for keys are able to return msgpack data.
+
+        :param path: None, tuple or string
+
+        """
+        if not MSGPACK_AVAILABLE:
+            return False
+        # path
```
chekunkov: oops, meaningless comment, fixed in fe134f0

PR #32, opened 2016-10-18
Pytest and vcrpy to improve sh.hubstorage tests

vshlapakov:
Please, review.


vshlapakov:
@chekunkov All the hubstorage tests were converted to pytest, and works perfectly with/without VCR.py. Last step is to commit all the locally generated cassetes and enable tests for sh.hs back. But these cassetes take a lot of space, about 6.5MB, so after the commit it will be very hard to review the changes itself; so I propose to review the changes, fix and merge it, and add the cassetes & enable the tests in other PR.

vshlapakov:
@chekunkov Last changes:
- reworked existing fixtures, used vcr decorator to have per-session cassettes
- found a culprit of high size of the cassettes (test_project.py:test_samples test), changed it a bit, now the cassette takes 200KB instead of 2+MB
- gzip serializer for vcr cassettes
- py.test option `--update-cassettes` to run tests with real services and update cassettes

The listed things helped to reduce cassettes size to ~800KB, wonder if it's ok. Most cassettes are already small enough, but probably we could win some space by rewriting tests with largest cassettes.
```
220K    cassetes/test_project/test_samples.gz
68K     cassetes/test_system/test_succeed_with_close_reason.gz
64K     cassetes/test_system/test_succeed_without_close_reason.gz
64K     cassetes/test_system/test_scraper_failure.gz
24K     cassetes/test_jobq/test_summary_countstart.gz
12K     cassetes/test_project/test_bulkdata.gz
```

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,175 @@
...
+my_vcr.register_serializer('gz', VCRGzipSerializer())
+my_vcr.serializer = 'gz'
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--update-cassettes", action="store_true", default=False,
+        help="make tests with real services instead of vcr cassettes")
+
+import shutil
```
chekunkov:
```
Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports

You should put a blank line between each group of imports.
```

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,175 @@
...
+    # otherwise it uses current function name (setup_vcrpy) for all tests
+    # other option is to add vcr decorator to each test separately
+    cassette_name = '{}/{}.gz'.format(
+        request.function.__module__.split('.')[-1],
+        request.function.__name__
+    )
+    with my_vcr.use_cassette(cassette_name):
+        remove_all_jobs(hsproject)
+        yield
+        remove_all_jobs(hsproject)
```
chekunkov: I think there's no need to remove all jobs after the test case, it's only needed to prepare environment for starting test case

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,175 @@
...
+def setup_vcrpy(request, hsproject):
+    # generates names like "test_module/test_function.yaml"
+    # otherwise it uses current function name (setup_vcrpy) for all tests
+    # other option is to add vcr decorator to each test separately
+    cassette_name = '{}/{}.gz'.format(
+        request.function.__module__.split('.')[-1],
+        request.function.__name__
+    )
+    with my_vcr.use_cassette(cassette_name):
+        remove_all_jobs(hsproject)
```
chekunkov: Maybe we can reduce size of the cassettes even more by excluding this setup if we are using cassettes? It's needed only when you run tests against live servers and brings no value while using requests recorded in the cassettes. E.g. we can do this cleanup only `if config.option.update_cassettes:`

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,175 @@
...
+
+my_vcr = vcr.VCR(cassette_library_dir=VCR_CASSETES_DIR, record_mode='once')
+my_vcr.register_serializer('gz', VCRGzipSerializer())
+my_vcr.serializer = 'gz'
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--update-cassettes", action="store_true", default=False,
+        help="make tests with real services instead of vcr cassettes")
```
chekunkov: What about adding another option that doesn't drop existing cassettes, just optionally ignores them? E.g. to check tests with real services, but not override existing cassettes without the need

tests/hubstorage/test_collections.py
```diff
@@ -8,123 +8,128 @@
...
+
+
+def test_errors(hscollection):
+    with pytest.raises(KeyError):
+        hscollection.get('does_not_exist')
+    args = [
+        {'foo': 42},
+        {'_key': []},
+        {'_key': 'large_test', 'value': 'x' * 1024 ** 2},
+    ]
```
chekunkov: is it possible to use `@pytest.mark.parametrize` instead?

chekunkov:
Don't forget to add vcrpy to requirements-test.txt

tests/hubstorage/conftest.py
vshlapakov: Oh, I missed this one, sorry!

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+        # or clear existing records, so its size will always only grow
+        if os.path.exists(VCR_CASSETES_DIR):
+            shutil.rmtree(VCR_CASSETES_DIR)
+    elif config.option.ignore_cassettes:
+        # simple hack to just ignore vcr cassettes:
+        # - all record_mode means recording new interactions + no replay
+        # - before_record returning None means skipping all the requests
+        global my_vcr
+        my_vcr.record_mode = 'all'
+        my_vcr.before_record_request = lambda request: None
```
chekunkov: 👍

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+    return str(hsproject.ids.spider(TEST_SPIDER_NAME, create=1))
+
+
+@my_vcr.use_cassette()
+@pytest.fixture(scope='session')
+def hscollection(hsproject):
+    collection = get_test_collection(hsproject)
+    clean_collection(collection)
+    yield collection
+    clean_collection(collection)
```
chekunkov: cleaning after the test is not necessary

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+@my_vcr.use_cassette()
+@pytest.fixture(autouse=True, scope='session')
+def setup_session(hsclient, hsproject, hscollection):
+    set_testbotgroup(hsproject)
+    remove_all_jobs(hsproject)
+    yield
+    remove_all_jobs(hsproject)
+    unset_testbotgroup(hsproject)
```
chekunkov: remove_all_jobs and unset_testbotgroup are not necessary after the test

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+    return hsclient.get_project(TEST_PROJECT_ID)
+
+
+@my_vcr.use_cassette()
+@pytest.fixture(scope='session')
+def hsspiderid(hsproject):
+    return str(hsproject.ids.spider(TEST_SPIDER_NAME, create=1))
+
+
+@my_vcr.use_cassette()
```
chekunkov: doubt that you need vcr here... get_test_collection doesn't produce requests, cleanup makes sense only when it's running with real services, like in https://github.com/scrapinghub/python-scrapinghub/pull/32/files#diff-5a347e09fdac148d5d13fac9f567fc37R121

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+    clean_collection(collection)
+    yield collection
+    clean_collection(collection)
+
+
+@my_vcr.use_cassette()
+@pytest.fixture(autouse=True, scope='session')
+def setup_session(hsclient, hsproject, hscollection):
+    set_testbotgroup(hsproject)
+    remove_all_jobs(hsproject)
```
chekunkov: do set_testbotgroup and remove_all_jobs make sense when running from vcrpy cassettes? if not - good candidate for vcr removal and condition like in https://github.com/scrapinghub/python-scrapinghub/pull/32/files#diff-5a347e09fdac148d5d13fac9f567fc37R121

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+
+@pytest.fixture(scope='session')
+def hsproject(hsclient):
+    return hsclient.get_project(TEST_PROJECT_ID)
+
+
+@my_vcr.use_cassette()
+@pytest.fixture(scope='session')
+def hsspiderid(hsproject):
+    return str(hsproject.ids.spider(TEST_SPIDER_NAME, create=1))
```
chekunkov: same as https://github.com/scrapinghub/python-scrapinghub/pull/32/files#r85515418

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+    for k in list(hsproject.settings.keys()):
+        if k != 'botgroups':
+            del hsproject.settings[k]
+    hsproject.settings.save()
+
+    # Cleanup JobQ
+    jobq = hsproject.jobq
+    for queuename in ('pending', 'running', 'finished'):
+        info = {'summary': [None]}  # poor-guy do...while
+        while info['summary']:
```
chekunkov: btw, it's not Pythonic :P 

```
while True:
    info = jobq.summary(queuename)
    if not info['summary']:
        break
    # ...
```

vshlapakov: Oh, sure! I just moved the function as-is and didn't notice it, thanks

tests/hubstorage/conftest.py
chekunkov: ok, after some thinking - let's keep it, minor thing and we cannot guarantee the same spider id in dev environment

tests/hubstorage/conftest.py
```diff
@@ -0,0 +1,189 @@
...
+my_vcr.serializer = 'gz'
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--update-cassettes", action="store_true", default=False,
+        help="test with real services rewriting existing vcr cassettes")
+    parser.addoption(
+        "--ignore-cassettes", action="store_true", default=False,
+        help="test with real services skipping existing vcr cassettes")
```
chekunkov: btw, this has to be added to the `tests/conftest.py`, otherwise it doesn't work with `pytest tests`

PR #34, opened 2016-11-03
Avoid races for hubstorage frontier tests

vshlapakov:
Looks like there're races in sh.hubstorage.frontier tests, it's relatively easy to reproduce by rerunning the Travis job (I can't reproduce it locally)
https://travis-ci.org/scrapinghub/python-scrapinghub/jobs/172664296

After checking internals, my guess is that as `Batchuploader` works in a separate thread trying to upload next messages batch from queue, and `frontier.flush()` operation waits for the queue being empty (by doing `queue.join()`),  there's a probability that on context switch the queue is empty, but `Batchuploader` hasn't called callback [yet](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/hubstorage/batchuploader.py#L120), and the `frontier.newcounter` is not updated yet. In this case simple short delay should fix this, at least I wasn't able to reproduce the issue after the fix.

Could you please confirm/disapprove my finding please?

### DIFF
--- a/scrapinghub/hubstorage/batchuploader.py
+++ b/scrapinghub/hubstorage/batchuploader.py
@@ -50,6 +50,8 @@ class BatchUploader(object):
     def create_writer(self, url, start=0, auth=None, size=1000, interval=15,
                       qsize=None, content_encoding='identity',
                       maxitemsize=1024 ** 2, callback=None):
+        # callback shouldn't try to inject more items in the queue
+        # otherwise it can lead to deadlock on _checkpoint step
         assert not self.closed, 'Can not create new writers when closed'
         auth = xauth(auth) or self.client.auth
         w = _BatchWriter(url=url,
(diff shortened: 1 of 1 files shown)


chekunkov:
@vshlapakov If there's an expectation that after a flush() counter will have a correct new value - IMO it's a bug and should be resolved by moving 

```
for _ in range(qiter.count):
    q.task_done()
```

after callback invocation [here](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/hubstorage/batchuploader.py#L130-L136).

vshlapakov:
@chekunkov I'm not quite sure that there's such expectation, the original [PR](https://github.com/scrapinghub/python-hubstorage/pull/12) doesn't contain anything related to it. But from my pov with your change the logic would be more consistent, let's improve it.
@dangra Any objections ?

chekunkov:
callbacks are only used in frontier afaik, operation in callback should be quick (parse json, increment count) - I don't think there are any downsides

dangra:
> callbacks are only used in frontier afaik, operation in callback should be quick (parse json, increment count) - I don't think there are any downsides

As long as callbacks doesn't try to inject more items in the queue it will work, otherwise it deadlocks.

vshlapakov:
@chekunkov @dangra Any pushback against the changes? I think it makes sense to drop a comment to avoid adding new elements in the queue with callbacks and merge the PR.

chekunkov:
+1 to drop the comment and merge. I can hardly imagine reasons to store more items in the queue after successful POST request - writers should not generate items.

PR #37, opened 2016-12-06
Revive NEWS file from GIT history

vshlapakov:
Revive NEWS file based on GIT history.

Review, please.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,9 +1,117 @@
 python-scrapinghub release notes
 ================================
 
+Version 1.9.0
+-------------
+(release on Nov 2nd, 2016)
+
+- python-hubstorage merged into python-scrapinghub
+- all tests are improved and rewritten with py.test
+- hubstorage tests use vcrpy cassettes, work faster and don't require any external services to run
+
+python-hubstorage is going to be considered deprecated,
+its next version will contain a deprecation warning and a proposal
+to use python-scrapinghub >=1.9.0 instead.
+
+Version 1.8.0
+-------------
+(release on July 29th, 2016)
+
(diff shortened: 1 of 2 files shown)


NEWS
```diff
@@ -1,9 +1,117 @@
 python-scrapinghub release notes
 ================================
 
+Version 1.9.0
+-------------
+(release on Nov 2nd, 2016)
```
chekunkov: why not to include date in the header, seems to be common practice

https://github.com/scrapy/scrapy/blob/master/docs/news.rst
https://github.com/kennethreitz/requests/blob/master/HISTORY.rst

I like requests approach `<version> (<date>)` the most - short and clean

NEWS
```diff
@@ -1,9 +1,117 @@
...
+
+- python-hubstorage merged into python-scrapinghub
+- all tests are improved and rewritten with py.test
+- hubstorage tests use vcrpy cassettes, work faster and don't require any external services to run
+
+python-hubstorage is going to be considered deprecated,
```
chekunkov: can we add a link to python-hubstorage?

NEWS
redapple: Note that starting with a digit seems to cause trouble with generated link (for Sphinx at least)
e.g for 2.12.3 (2016-12-01), the link is
`http://docs.python-requests.org/en/master/community/updates/#id1`

I "fixed" it in Scrapy by prepending "Scrapy " (https://github.com/scrapy/scrapy/pull/2294)

chekunkov: that's strange @redapple we'll change the headers later if starting digit will stand in our way :)

PR #38, opened 2016-12-14
SH Python client update

vshlapakov:
2nd attempt to merge, simplify and fix existing SH python clients, based on composition approach.

Please ignore codecov warnings: the PR contains a lot of tests - it should be enough, but they are disabled cause based on VCR.py cassettes that I'll add in a separate PR shortly once this PR is approved and merged.


dangra:
Looks good @vshlapakov

vshlapakov:
@dangra Ok, now it's ready for careful review, I've just finished with VCR cassettes but will add it in a separate PR to reduce amount of the changes diff.
VCP.py results

scrapinghub/client.py
```diff
@@ -0,0 +1,350 @@
...
+    def __init__(self, client, projectid):
+        self.id = projectid
+        self._client = client
+
+        # sub-resources
+        self.jobs = Jobs(client, projectid)
+        self.spiders = Spiders(client, projectid)
+
+        # proxied sub-resources
+        self.activity = Activity(client._hsclient, projectid)
```
chekunkov: `proj.activity.list()` returns generator. I think we shouldn't leave non-proxied objects in the new client, it may cause confusion

chekunkov:
```
In [38]: proj.activity.add({'job': 'xxx', 'user': 'xxx', 'event': 'hey-ho'})
Out[38]: <generator object jldecode at 0x106eee730>
```

don't like ^ as well

scrapinghub/client.py
```diff
@@ -0,0 +1,349 @@
...
+from scrapinghub import HubstorageClient
+
+from .hubstorage.resourcetype import DownloadableResource
+from .hubstorage.resourcetype import ItemsResourceType
+
+# scrapinghub.hubstorage classes to use as-is
+from .hubstorage.activity import Activity
+from .hubstorage.frontier import Frontier
+from .hubstorage.job import JobMeta
+from .hubstorage.project import Reports
```
chekunkov: don't forget to remove import

scrapinghub/client.py
vshlapakov: Well noticed, I'll fix it 👍

README_client.rst
```diff
@@ -0,0 +1,480 @@
...
+-------
+
+Project instance has ``jobs`` field to work with the project jobs.
+
+Jobs instance is described well in ``Jobs`` section below.
+
+For example, to schedule a spider run (it returns a job object)::
+
+    >>> project.jobs.schedule('spider1', arg1='val1')
+    <scrapinghub.client.Job at 0x106ee12e8>>
```
chekunkov: @vshlapakov wouldn't it be much nicer for future to have dedicated parameter for spider arguments instead of assuming that all unknown kwargs in the `schedule` methods are spider args?

vshlapakov: @chekunkov Not sure tbh: it's less error-prone, but spider args is the most frequent parameter here, it can be annoying to always provide it explicitly `spider_args={..}`; I think we could keep it as-is for now.

scrapinghub/client.py
```diff
@@ -0,0 +1,370 @@
...
+        """ Iterate over jobs collection for a given set of params.
+        FIXME the function returns a list of dicts, not a list of Job's
+        """
+        if self.spider:
+            params['spider'] = self.spider.name
+        return self._project.jobq.list(**params)
+
+    def schedule(self, spidername=None, **params):
+        if not spidername and not self.spider:
+            raise APIError('Please provide spidername')
```
chekunkov: I'd say it should be a `ValueError`, `APIError` now mixes errors from Connection (HTTP errors?), errors from new client. IMO:

- `ValueError` is what needed if method receives unexpected argument (or doesn't receive expected)
- new client deserves new base Exception class, new exceptions like DuplicateJobError should inherit from new base class - let's keep APIError for legacy Connection 
- Most of 4xx responses should be covered with specific exceptions classes, like 400 should result in `ValidationError` or `InvalidUsage`, 404 in generic `NotFound`. they also should store either original `requests.exceptions.HTTPError` or response object - to be able to get body and check why things failed.
- 5xx errors should be raised as is I think, we just need to be sure it's always requests.exceptions.HTTPError

scrapinghub/client.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+class Collections(_Proxy):
+
+    def __init__(self, *args, **kwargs):
+        super(Collections, self).__init__(*args, **kwargs)
+        self._proxy_methods([
+            'count', 'get', 'set', 'delete', 'create_writer',
+            '_validate_collection',
+        ])
```
chekunkov: I don't think we need to proxy these methods, they are only used internally from `hubstorage.collectionsrt.Collection`

scrapinghub/client.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def new_store(self, colname):
```
chekunkov: IMO `new_` should be replaced with `get_` for all the following methods. it's not new collection if it was already created and `project.collections.get_store('foo')` looks a bit more relevant here.

scrapinghub/client.py
```diff
@@ -0,0 +1,370 @@
...
+        return Collection(self._client, self, coltype, colname)
+
+
+class Collection(object):
+
+    def __init__(self, client, collections, coltype, colname):
+        self._client = client
+        self._origin = _Collection(coltype, colname, collections._origin)
+        proxy_methods(self._origin, self, [
+            'create_writer', 'get', 'set', 'delete', 'count',
```
chekunkov: one note about `get` method - in Hubstorage client if you don't pass key to get - it returns an iterator over the entire collection. literally the same thing as `iter`. I think it makes more sense to make `key` parameter required for `get` method.

chekunkov: I'm not sure we should proxy create_writer here, it's not clear how it can be used, it may be some legacy code.

scrapinghub/utils.py
```diff
@@ -4,22 +4,38 @@
...
-class DuplicateJobError(APIError):
+    def __init__(self, origin):
+        self.origin = origin
+        message = _get_http_error_msg(origin)
+        super(ScrapinghubAPIError, self).__init__(message)
+
+
+def _get_http_error_msg(exc):
+    try:
+        return exc.response.json()
```
chekunkov: you should check if origin is HTTPError

scrapinghub/client.py
vshlapakov: Agree to drop all of them except `_validate_collection` - it's reused in `new_collection` method (and in other `new_` methods as a consequence)

scrapinghub/utils.py
```diff
@@ -4,22 +4,38 @@
 from six import string_types
 from requests.exceptions import HTTPError
 
-from scrapinghub import APIError
 
+class ScrapinghubAPIError(Exception):
 
-class DuplicateJobError(APIError):
+    def __init__(self, origin):
```
chekunkov: should be possible to raise `ScrapinghubAPIError` without origin and with some custom error message

scrapinghub/client.py
vshlapakov: Yes, agree on both points, thanks

scrapinghub/utils.py
vshlapakov: Already done, please update the page

scrapinghub/client.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def __init__(self, client, projectid):
+        self.projectid = projectid
+        self._client = client
+
+    def get(self, spidername, **params):
+        project = self._client._hsclient.get_project(self.projectid)
+        spiderid = project.ids.spider(spidername, **params)
+        if spiderid is None:
+            raise ValueError("Spider {} doesn't exist.".format(spidername))
```
chekunkov: I'd say NotFound

scrapinghub/utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+def wrap_kwargs(fn, kwargs_fn):
+    """Tiny wrapper to prepare modified version of function kwargs"""
+    def wrapped(*args, **kwargs):
+        kwargs = kwargs_fn(kwargs)
+        return fn(*args, **kwargs)
+    return wrapped
+
+
+def proxy_methods(origin, successor, methods, wrapper=wrap_http_errors):
```
chekunkov: wrapper=wrap_http_errors is not needed here because requests are already wrapped in hs client

vshlapakov: Yeah, sorry, forgot to drop it

scrapinghub/client.py
```diff
@@ -0,0 +1,350 @@
...
+        return self._client._hsclient.get_project(self.projectid)
+
+    def count(self, **params):
+        if self.spider:
+            params['spider'] = self.spider.name
+        return next(self._hsproject.jobq.apiget(('count',), params=params))
+
+    def iter(self, **params):
+        """ Iterate over jobs collection for a given set of params.
+        FIXME the function returns a list of dicts, not a list of Job's
```
vshlapakov: Agreed that `iter` returns jobs summary, not jobs.
Let's return to the question when there's any real need for it

README_client.rst
chekunkov: @vshlapakov if we decide to add spider_args later that would be a backwards incompatible change if someone accidentally has spider argument `spider_args`. another problem is that parameters like `units` or `priority` or `spider_settings` are not distinguishable from any other spider argument. by having separate kwarg for spider args we can guarantee that if we add new parameters to schedule endpoint they would not clash with spider args that clients use

scrapinghub/client.py
vshlapakov: Agree to keep create_writer - could be useful for writing to collection in batches

README_client.rst
vshlapakov: That's correct, I really have nothing against `spider_args` except this little doubt about convenience. Alright, let's better change it now than later

README.rst
```diff
@@ -31,8 +31,483 @@ response time and improved bandwidth usage::
...
+      {'id': 'spider2', 'tags': [], 'type': 'manual', 'version': '123'}
+    ]
+
+To select a particular spider to work with::
+
+    >>> spider = project.spiders.get('spider2')
+    >>> spider
+    <scrapinghub.client.Spider at 0x106ee3748>
+    >>> spider.id
+    2
```
chekunkov: btw, note that `spider.id` is not consistent with `job.id`

scrapinghub/client.py
```diff
@@ -203,6 +195,17 @@ def _extract_spider_id(self, params):
             raise ValueError('Please use same spider id')
         return spiderid
 
+    def update_tags(self, add=None, remove=None, spidername=None):
+        spidername = spidername or (self.spider.name if self.spider else None)
+        if not spidername:
+            raise ValueError('Please provide spidername')
```
vshlapakov: Sure, it covers the case when we work with project jobs (`project.jobs`), that's a trick to have the same logic for project jobs and spider jobs

chekunkov: @vshlapakov this validation check starts worrying me, shouldn't it be done on the dash side?

scrapinghub/client.py
```diff
@@ -247,16 +247,16 @@ def start(self, **params):
...
     def update(self, **params):
-        return self._project.jobq.update(self, **params)
+        return next(self._project.jobq.update(self, **params))['prevstate']
 
     def cancel(self):
         self._project.jobq.request_cancel(self)
 
     def finish(self, **params):
-        return self._project.jobq.finish(self, **params)
+        return next(self._project.jobq.finish(self, **params))['prevstate']
```
vshlapakov: Maybe I'm missing something, but if a) response is an empty list and b) `finish()` returns iterator wrapping it, then `next()` operation will raise StopIteration anyway. And if response is not empty, it will return a corresponding `prevstate`.
```
>>> a = iter([1])
>>> str(next(a))
  File "<stdin>", line 1, in <module>
StopIteration
```

scrapinghub/client.py
vshlapakov: @chekunkov Makes sense to me, I'd move the checks to Dash then, maybe someday we add scheduling a default/last-launched spider, and updating tags for a whole project will also be allowed.

For notes now Dash validates spidername on scheduling jobs pretty good, but it expects `spider` param while client `schedule()` accepts `spidername`, I can rename it in the client or we should map it in the Dash:
```
{"spider":["This field is required."],"status":"error","message":"spider: This field is required."}
```
As for updating tags without specifying a spider - it's less clear, there should be a check:
```
{"message": "No query filters provided", "status": "badrequest"}
```

scrapinghub/client.py
chekunkov: yes, you understand the behaviour correctly. is that expected by user though?

scrapinghub/client.py
chekunkov: ^ worth discussing with @michalmo 

if changes are hard or undesirable on the dash side - we can stick with ValueError for now, just move it to a reusable method.

renaming `spidername` to `spider` in the client could be confusing, I'd expect `spider` to be an instance of `Spider`
```

```

scrapinghub/client.py
vshlapakov: Of course not, I see what do you mean now, will fix it, thanks!

vshlapakov: Moved the changes into a separate PR for simplicity https://github.com/scrapinghub/python-scrapinghub/pull/46.

scrapinghub/exceptions.py
```diff
@@ -62,6 +63,11 @@ def wrapped(*args, **kwargs):
             elif 400 <= status_code < 500:
                 raise ScrapinghubAPIError(http_error=exc)
             raise
+        except APIError as exc:
+            msg = exc.args[0]
+            if 'format must be either json or jl' in msg:
```
vshlapakov: @chekunkov Did you mean something like that https://github.com/scrapinghub/python-scrapinghub/pull/47? 
(Moved the changes to a separate PR for simplicity.)

PR #46, opened 2017-01-31
Updating job state: return previous state

vshlapakov:
Please, review.

### DIFF
--- a/scrapinghub/client.py
+++ b/scrapinghub/client.py
@@ -237,19 +237,23 @@ class Job(object):
         self.metadata.expire()
 
     def start(self, **params):
-        return self._project.jobq.start(self, **params)
+        return self.update(state='running', **params)
+
+    def finish(self, **params):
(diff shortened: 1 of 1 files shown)


scrapinghub/client.py
```diff
@@ -237,20 +237,24 @@ def purge(self):
...
+    def delete(self, **params):
+        return self.update(state='deleted', **params)
 
     def update(self, **params):
-        return self._project.jobq.update(self, **params)
+        try:
+            job = next(self._project.jobq.update(self, **params))
+            return job['prevstate']
+        except StopIteration:
+            raise ValueError("Job {} doesn't exist".format(self.key))
```
chekunkov: why ValueError?

vshlapakov: User can pass any non-existing job key, for me `ValueError` or `KeyError` should fit, do you have better idea in mind?

PR #47, opened 2017-02-01
Add type field to APIError for better matching errors

vshlapakov:
Raised by [the comment](https://github.com/scrapinghub/python-scrapinghub/pull/38#discussion_r98424409), should simplify matching and handling exceptions by type in the new python client. Review, please.

### DIFF
--- a/scrapinghub/legacy.py
+++ b/scrapinghub/legacy.py
@@ -97,7 +97,8 @@ class Connection(object):
         try:
             base_path = self.API_METHODS[method]
         except KeyError:
-            raise APIError("Unknown method: {0}".format(method))
+            raise APIError("Unknown method: {0}".format(method),
+                           _type=APIError.ERR_VALUE_ERROR)
         else:
             path = "{0}.{1}".format(base_path, format)
             return urljoin(self.url, path)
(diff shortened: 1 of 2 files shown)


scrapinghub/legacy.py
```diff
@@ -97,7 +97,8 @@ def _build_url(self, method, format):
         try:
             base_path = self.API_METHODS[method]
         except KeyError:
-            raise APIError("Unknown method: {0}".format(method))
+            raise APIError("Unknown method: {0}".format(method),
+                           _type=APIError.ERR_INVALID_USAGE)
```
chekunkov: seems to be value error though

scrapinghub/legacy.py
```diff
@@ -145,7 +147,8 @@ def _decode_response(self, response, format, raw):
                 if data['status'] == 'ok':
                     return data
                 elif data['status'] in ('error', 'badrequest'):
-                    raise APIError(data['message'])
+                    raise APIError(data['message'],
+                                   _type=APIError.ERR_VALUE_ERROR)
```
chekunkov: and here it's more like 400 or invalid usage

vshlapakov: Right, agree

PR #48, opened 2017-02-07
Improvements and fixes for new python client

vshlapakov:
The PR is created to cover the following minor things found after testing:
- [x] fix inconsistencies between `project.id - spider.id - job.key` (use `key` or `id` everywhere)
- [x] collection.set() should return None instead of a generator
- [x] simplify and unify using HS `filter` param 
- [x] activity.add(): add client validation for project id
- [x] collections.iter() should iterate through collections
- [x] unify errors when apikey is wrong (now it's `InvalidUsage` or `ScrapinghubAPIError`)
- [x] handle HTTP 401 from JobQ separately with providing a good message instead of html

Other issues added after next step of testing at February 8th:
- [x] collection.delete() should return None instead of a generator
- [x] add validation for collection.delete input
- [x] validate that update_tags method has any input

Also there's a new `spider.update_tags` method that will be added on server-side soon.

Each of the points will be implemented in a separate commit(s) to simplify reviewing.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+addopts = --doctest-modules --doctest-glob='scrapinghub/*.py'
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --doctest-modules --doctest-glob='scrapinghub/*.py'
+addopts = --doctest-glob='scrapinghub/*.py'
--- a/scrapinghub/client.py
+++ b/scrapinghub/client.py
@@ -391,6 +391,16 @@ class Jobs(object):
             params['spider'] = self.spider.name
         return self._project.jobq.list(**params)
 
+    def list(self, **params):
+        """Convenient shortcut to list iter results.
+
+        Please note that list() method can use a lot of memory and for a large
+        amount of jobs it's recommended to iterate through it via iter()
+        method (all params and available filters are same for both methods).
+
+        """
+        return list(self.iter(**params))
+
     def schedule(self, spidername=None, **params):
         """Schedule a new job and returns its jobkey.
 
@@ -762,12 +772,19 @@ class _Proxy(object):
         """Modify iter() params on-the-fly."""
         return format_iter_filters(params)
 
+    def list(self, *args, **kwargs):
+        return list(self.iter(*args, **kwargs))
+
 
 class Logs(_Proxy):
     """Representation of collection of job logs.
 
     Not a public constructor: use :class:`Job` instance to get a :class:`Logs`
     instance. See :attr:`Job.logs` attribute.
+
+    Please note that list() method can use a lot of memory and for a large
+    amount of logs it's recommended to iterate through it via iter() method
+    (all params and available filters are same for both methods).
 
     Usage:
 
@@ -776,9 +793,14 @@ class Logs(_Proxy):
         >>> job.logs.iter()
         <generator object mpdecode at 0x10f5f3aa0>
 
+    - iterate through first 100 log entries and print them::
+
+        >>> for log in job.logs.iter(count=100):
+        >>> ... print(log)
+
     - retrieve a single log entry from a job::
 
-        >>> list(job.logs.iter(count=1))
+        >>> job.logs.list(count=1)
         [{
             'level': 20,
             'message': '[scrapy.core.engine] Closing spider (finished)',
@@ -788,7 +810,7 @@ class Logs(_Proxy):
     - retrive logs with a given log level and filter by a word
 
         >>> filters = [("message", "contains", ["logger"])]
-        >>> list(job.logs.iter(level='WARNING', filter=filters))
+        >>> job.logs.list(level='WARNING', filter=filters)
         [{
             'level': 30,
             'message': 'Some warning message',
@@ -831,6 +853,10 @@ class Items(_Proxy):
     Not a public constructor: use :class:`Job` instance to get a :class:`Items`
     instance. See :attr:`Job.items` attribute.
 
+    Please note that list() method can use a lot of memory and for a large
+    amount of items it's recommended to iterate through it via iter() method
+    (all params and available filters are same for both methods).
+
     Usage:
 
     - retrieve all scraped items from a job::
@@ -838,20 +864,29 @@ class Items(_Proxy):
         >>> job.items.iter()
         <generator object mpdecode at 0x10f5f3aa0>
 
+    - iterate through first 100 items and print them::
+
+        >>> for log in job.logs.iter(count=100):
+        >>> ... print(log)
+
     - retrieve items with timestamp greater or equal to given timestamp
       (item here is an arbitrary dictionary depending on your code)::
 
-        >>> list(job.items.iter(startts=1447221694537))
-        {'name': ['Some custom item'],
-         'url': 'http://some-url/item.html',
-         'size': 100000}
+        >>> job.items.list(startts=1447221694537)
+        [{
+            'name': ['Some custom item'],
+            'url': 'http://some-url/item.html',
+            'size': 100000,
+        }]
 
     - retrieve 1 item with multiple filters:
         >>> filters = [("size", ">", [30000]), ("size", "<", [40000])]
-        >>> list(job.items.iter(count=1, filter=filters))
-        {'name': ['Some other item'],
-         'url': 'http://some-url/other-item.html',
-         'size': 50000}
+        >>> job.items.list(count=1, filter=filters)
+        [{
+            'name': ['Some other item'],
+            'url': 'http://some-url/other-item.html',
+            'size': 50000,
+        }]
     """
 
     def _modify_iter_params(self, params):
@@ -873,6 +908,10 @@ class Requests(_Proxy):
     Not a public constructor: use :class:`Job` instance to get a
     :class:`Requests` instance. See :attr:`Job.requests` attribute.
 
+    Please note that list() method can use a lot of memory and for a large
+    amount of requests it's recommended to iterate through it via iter()
+    method (all params and available filters are same for both methods).
+
     Usage:
 
     - retrieve all requests from a job::
@@ -888,7 +927,7 @@ class Requests(_Proxy):
 
     - retrieve single request from a job::
 
-        >>> list(job.requests.iter(count=1))
+        >>> job.requests.list(count=1)
         [{
         'duration': 354,
         'fp': '6d748741a927b10454c83ac285b002cd239964ea',
@@ -910,6 +949,10 @@ class Samples(_Proxy):
     Not a public constructor: use :class:`Job` instance to get a
     :class:`Samples` instance. See :attr:`Job.samples` attribute.
 
+    Please note that list() method can use a lot of memory and for a large
+    amount of samples it's recommended to iterate through it via iter()
+    method (all params and available filters are same for both methods).
+
     Usage:
 
     - retrieve all samples from a job::
@@ -919,7 +962,7 @@ class Samples(_Proxy):
 
     - retrieve samples with timestamp greater or equal to given timestamp::
 
-        >>> list(job.samples.iter(startts=1484570043851))
+        >>> job.samples.list(startts=1484570043851)
         [[1484570043851, 554, 576, 1777, 821, 0],
          [1484570046673, 561, 583, 1782, 821, 0]]
     """
@@ -931,6 +974,10 @@ class Activity(_Proxy):
     Not a public constructor: use :class:`Project` instance to get a
     :class:`Activity` instance. See :attr:`Project.activity` attribute.
 
+    Please note that list() method can use a lot of memory and for a large
+    amount of activities it's recommended to iterate through it via iter()
+    method (all params and available filters are same for both methods).
+
     Usage:
 
     - get all activity from a project::
@@ -940,7 +987,7 @@ class Activity(_Proxy):
 
     - get only last 2 events from a project::
 
-        >>> list(p.activity.iter(count=2))
+        >>> p.activity.list(count=2)
         [{'event': 'job:completed', 'job': '123/2/3', 'user': 'jobrunner'},
          {'event': 'job:cancelled', 'job': '123/2/3', 'user': 'john'}]
     """
@@ -1026,13 +1073,14 @@ class Collection(object):
 
     - iterate iterate over _key & value pair::
 
-        >>> list(foo_store.iter())
-            [{'_key': '002d050ee3ff6192dcbecc4e4b4457d7',
-              'value': '1447221694537'}]
+        >>> for elem in foo_store.iter(count=1)):
+        >>> ... print(elem)
+        [{'_key': '002d050ee3ff6192dcbecc4e4b4457d7',
+            'value': '1447221694537'}]
 
     - filter by multiple keys, only values for keys that exist will be returned::
 
-        >>> list(foo_store.iter(key=['002d050ee3ff6192dcbecc4e4b4457d7', 'blah']))
+        >>> foo_store.list(key=['002d050ee3ff6192dcbecc4e4b4457d7', 'blah'])
         [{'_key': '002d050ee3ff6192dcbecc4e4b4457d7', 'value': '1447221694537'}]
 
     - delete an item by key::
@@ -1053,6 +1101,15 @@ class Collection(object):
         for method in ['iter', 'iter_raw_json']:
             wrapped = wrap_kwargs(getattr(self, method), format_iter_filters)
             setattr(self, method, wrapped)
+
+    def list(self, *args, **kwargs):
+        """Convenient shortcut to list iter results.
+
+        Please note that list() method can use a lot of memory and for a large
+        amount of elements it's recommended to iterate through it via iter()
+        method (all params and available filters are same for both methods).
+        """
+        return list(self.iter(*args, **kwargs))
 
     def get(self, key, *args, **kwargs):
         """Get item from collection by key.
--- a/scrapinghub/client.py
+++ b/scrapinghub/client.py
@@ -1018,3 +1018,10 @@ class Collection(object):
         if key is None:
             raise ValueError("key cannot be None")
         return self._origin.get(key, *args, **kwargs)
+
+    def set(self, *args, **kwargs):
+        """Set item to collection by key.
+
+        The method returns None (original method returns an empty generator).
+        """
+        self._origin.set(*args, **kwargs)
--- a/scrapinghub/client.py
+++ b/scrapinghub/client.py
@@ -1,5 +1,7 @@
 import json
 
+from six import string_types
+
 from scrapinghub import APIError
 from scrapinghub import Connection as _Connection
 from scrapinghub import HubstorageClient as _HubstorageClient
@@ -1092,7 +1094,7 @@ class Collection(object):
         self._client = client
         self._origin = _Collection(coltype, colname, collections._origin)
         proxy_methods(self._origin, self, [
-            'create_writer', 'delete', 'count',
+            'create_writer', 'count',
             ('iter', 'iter_values'),
             ('iter_raw_json', 'iter_json'),
         ])
@@ -1127,3 +1129,15 @@ class Collection(object):
         The method returns None (original method returns an empty generator).
         """
         self._origin.set(*args, **kwargs)
+
+    def delete(self, _keys):
+        """Delete item(s) from collection by key(s).
+
+        The method returns None (original method returns an empty generator).
+        """
+        if (not isinstance(_keys, string_types) and
+                not(isinstance(_keys, (list, tuple)) and
+                    all(isinstance(key, string_types) for key in _keys))):
+            raise ValueError(
+                "You should provide a string key or a list of string keys")
+        self._origin.delete(_keys)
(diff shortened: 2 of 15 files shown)


scrapinghub/utils.py
```diff
@@ -83,3 +84,20 @@ def proxy_methods(origin, successor, methods):
...
+    """Format iter() filter param on-the-fly.
+
+    Support passing multiple filters at once as a list with tuples.
+    """
+    filters = params.get('filter')
+    if filters and isinstance(filters, list):
+        filter_data = []
+        for elem in params.pop('filter'):
+            if not isinstance(elem, (list, tuple)):
+                raise ValueError("Filter condition must be tuple or list")
```
chekunkov: I think we still should accept strings as a input format

scrapinghub/utils.py
```diff
@@ -83,3 +84,20 @@ def proxy_methods(origin, successor, methods):
...
+            filter_data.append(json.dumps(elem))
```
chekunkov: `elem` len is validated on the server side, right?

scrapinghub/client.py
```diff
@@ -284,15 +285,15 @@ class Spider(object):
 
         >>> project.spiders.get('spider2')
         <scrapinghub.client.Spider at 0x106ee3748>
-        >>> spider.id
+        >>> spider.key
         2
```
chekunkov: project.key = project id
job.key = '123/1/2'

spider.key == 2?

scrapinghub/client.py
```diff
@@ -200,7 +201,7 @@ class Project(object):
     """
 
     def __init__(self, client, projectid):
-        self.id = projectid
+        self.key = projectid
```
chekunkov: job.key is a string, shouldn't project key be a string as well?

scrapinghub/client.py
```diff
@@ -774,7 +780,17 @@ class Logs(_Proxy):
...
             'level': 20,
             'message': '[scrapy.core.engine] Closing spider (finished)',
-            'time': 1482233733976},
+            'time': 1482233733976,
+        }]
+
+    - retrive logs with a given log level and filter by a word
+
+        >>> filters = [("message", "contains", ["logger"])]
+        >>> list(job.logs.iter(level='WARNING', filter=filters))
```
chekunkov: I know I asked that before but maybe we should have a `entity.list()` method instead of `list(entity.iter())` pattern?

scrapinghub/client.py
```diff
@@ -1018,3 +1061,10 @@ def get(self, key, *args, **kwargs):
         if key is None:
             raise ValueError("key cannot be None")
         return self._origin.get(key, *args, **kwargs)
+
+    def set(self, *args, **kwargs):
```
chekunkov: hm, this method is still mentioned in the `proxy_methods(self._origin, self, [...])` call https://github.com/scrapinghub/python-scrapinghub/pull/48/files#diff-797cd9f74e9304996bef041260c04262R1044

scrapinghub/utils.py
vshlapakov: Yes, examples:
```
InvalidUsage: unable to parse filter parameter: Match spec must contain 3 values
InvalidUsage: unable to parse filter parameter: incorrect arguments for operator isnotempty
```

scrapinghub/client.py
vshlapakov: I've updated all the docstrings with the following model:
- projects 123 and 456
- project 123 has 2 spiders: (1, spider1) and (2, spider2)

scrapinghub/client.py
```diff
@@ -282,17 +285,17 @@ class Spider(object):
 
     Usage::
 
-        >>> project.spiders.get('spider2')
+        >>> spider = project.spiders.get('spider1')
         <scrapinghub.client.Spider at 0x106ee3748>
-        >>> spider.id
-        2
+        >>> spider.key
+        '1'
```
chekunkov: spider.id is 1 (or '1', whatever), but spider.key should be '123/1'

scrapinghub/client.py
chekunkov: @vshlapakov don't forget to ensure that key is str `self.key = str(projectid)`

and add line with `>>> project.key` output

scrapinghub/client.py
```diff
@@ -630,6 +632,8 @@ def update_tags(self, add=None, remove=None):
 
             >>> job.update_tags(add=['consumed'])
         """
+        if not (add or remove):
+            raise ValueError('Please provide tags to add or remove')
```
chekunkov: just in case, shouldn't it be handled by server side validation?

vshlapakov: Server responds with 200 for now, we may change it, but what's the point to do an empty request?

scrapinghub/client.py
vshlapakov: Ah, I misread your comment, sure

scrapinghub/client.py
chekunkov: but if it's empty - shouldn't api server return 400?

vshlapakov: Well, it makes sense, I don't mind to add it 👌  But I still don't see any harm to have client validation for it, there's no point to do update() call without data, do you think it's redundant?

chekunkov: yeah, redundancy is my main concern. when data is validated server side - you only need to change it in one place later

vshlapakov: Ok, convinced, I'll create a ticket for server-side changes and drop it from here

vshlapakov:
@chekunkov Ok, I think we've done with fixes/improvements, the PR deserves final review and merge.

scrapinghub/client.py
```diff
@@ -305,6 +306,16 @@ def __init__(self, client, projectid, spiderid, spidername):
...
         self._client = client
 
+    @wrap_http_errors
+    def update_tags(self, add=None, remove=None):
+        params = get_tags_for_update(add=add, remove=remove)
+        if not params:
+            return
+        path = 'v2/projects/{}/spiders/{}/tags'.format(self.projectid, self.id)
+        url = urljoin(self._client._connection.url, path)
+        response = self._client._connection._session.patch(url, json=params)
```
vshlapakov: @chekunkov I can't reuse existing `Connection` class in a better way, as there's no PATCH request type there and adding it doesn't look like a good idea. Adding similar system (`_get`/`_post`,`_build_url`) to the new `ScrapinghubClient` class also doesn't look appropriate now, do you agree that using something simple here makes sense for now? Later when we add some new functionality, we can move the part into a separate function(s).

chekunkov: Agree. Do we want to expose v2 api though?

scrapinghub/client.py
```diff
@@ -282,21 +288,34 @@ class Spider(object):
 
     Usage::
 
-        >>> project.spiders.get('spider2')
+        >>> spider = project.spiders.get('spider1')
         <scrapinghub.client.Spider at 0x106ee3748>
```
chekunkov: different from python interpreter output

```
>>> spider = project.spiders.get('localinfo')
>>> spider
<scrapinghub.client.Spider object at 0x10d6cb1d0>
```

I think you can simply remove `<scrapinghub.client.Spider at 0x106ee3748>`

scrapinghub/client.py
```diff
@@ -282,21 +288,34 @@ class Spider(object):
...
         >>> spider.name
-        spider2
+        'spider1'
     """
 
     def __init__(self, client, projectid, spiderid, spidername):
         self.projectid = projectid
-        self.id = spiderid
+        self.key = '{}/{}'.format(str(projectid), str(spiderid))
+        self.id = str(spiderid)
```
chekunkov: I find co-existance of id and key in Spider object a little bit confusing. Do we need them both? If so - why project and job object don't have both key and id. If id is needed only for internal use (e.g. to construct queries) - maybe we can make it private and remove from docs? If we expect users to use it - let's keep it public, but let's also be consistent.

scrapinghub/client.py
```diff
@@ -954,6 +1056,9 @@ def get_versioned_store(self, colname):
     def get_versioned_cached_store(self, colname):
         return self.get('vcs', colname)
 
+    def list(self):
+        return list(self._origin.apiget('list'))
```
chekunkov: let's add iter() for consistency?

chekunkov: the same for other classes that have list() method, but don't have iter() yet, e.g. Projects

scrapinghub/client.py
```diff
@@ -1018,3 +1138,22 @@ def get(self, key, *args, **kwargs):
...
         return self._origin.get(key, *args, **kwargs)
+
+    def set(self, *args, **kwargs):
+        """Set item to collection by key.
+
+        The method returns None (original method returns an empty generator).
+        """
+        self._origin.set(*args, **kwargs)
+
+    def delete(self, _keys):
```
chekunkov: why `_keys`? `keys` is not reserved or builtin

scrapinghub/client.py
```diff
@@ -1018,3 +1138,22 @@ def get(self, key, *args, **kwargs):
...
+        """Delete item(s) from collection by key(s).
+
+        The method returns None (original method returns an empty generator).
+        """
+        if (not isinstance(_keys, string_types) and
+                not(isinstance(_keys, (list, tuple)) and
+                    all(isinstance(key, string_types) for key in _keys))):
```
chekunkov:
```
if not isinstance(keys, (list, tuple)):
    keys = [keys]
if not all(isinstance(key, string_types) for key in keys)):
    raise ValueError()
```

scrapinghub/client.py
vshlapakov: I don't have any objections to expose it, but I can miss some details here. / cc @michalmo  

If we don't want to expose it for now, we should drop the spider's `list_tags`/`update_tags` from the client, or move the endpoints/add some routers.

scrapinghub/client.py
vshlapakov: Spider id is needed only for internal use like you mentioned, to avoid splitting the key all the time to construct queries and do some checks. Let's better make it private and remove from the docs, it makes sense to have it only for spider.

scrapinghub/client.py
vshlapakov: Renamed (initially was done for consistency with original method, but it's a not an argument)

scrapinghub/utils.py
```diff
@@ -41,10 +42,10 @@ def parse_job_key(jobkey):
     if len(parts) != 3:
         raise ValueError("Job key should consist of projectid/spiderid/jobid")
     try:
-        parts = map(int, parts)
+        map(int, parts)
     except ValueError:
         raise ValueError("Job key parts should be integers")
-    return JobKey(*parts)
+    return JobKey(*map(str, parts))
```
chekunkov: This change is not backwards compatible

chekunkov: Ah, seem like this entire file is something that has never been merged to the master, so probably we are good here... Correct me if I'm wrong

tests/client/conftest.py
```diff
@@ -69,7 +69,7 @@ def is_using_real_services(request):
 
 @pytest.fixture(scope='session')
 def client():
-    return ScrapinghubClient(auth=TEST_ADMIN_AUTH,
+    return ScrapinghubClient(apikey=TEST_ADMIN_AUTH,
```
chekunkov: @vshlapakov  just noticed you replaced `auth` with `apikey`. note that in some cases we pass colon separated auth string to the HubstorageClient, do you think this will never happen here? 

https://github.com/scrapinghub/scrapinghub-entrypoint-scrapy/blob/9e5c4f79f4fa2442cedf7e13185c7087961f6b46/sh_scrapy/env.py#L53-L53

scrapinghub/utils.py
vshlapakov: No, you're right, it's a new module, we should be good

tests/client/conftest.py
vshlapakov: Sorry for not being sound enough about the change, that's my bad.

Right, I think it will never happen, my args are the following:
- `Connection` works only with [apikey](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/legacy.py#L83), `HubstorageClient` [logic](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/legacy.py#L83) is more flexible, it also accepts `(user,pass)` as a tuple, or as a colon-separated auth string, so apikey looks like a common denominator for both clients, simple and power enough;
- apikey key looks like an enhancement over user/pass pair, and as last approach was dropped from `Connection` at some [point](https://github.com/scrapinghub/python-scrapinghub/commit/6655c2b67e8e21a95a5d85fd9694f1bc1b8cc40f#diff-5908bb1e4376e7f4b8bcf236a31c9487R52), I think we're not going to revise it in the nearest future.

Anyway let's revise it if you have something in mind.

chekunkov: my point is the following - users can add new client to the requirements and use it in Scrapy Cloud. when they do so, they would probably use auth from the environment - and that's where *have* to pass decoded `job_key:JWT_token` pair, they don't have an option to pass an apikey and if they pass that pair as an apikey - their requests to dash endpoint will fail with 401 because Connection doesn't expect such input and uses apikey parameter as a "user" part of auth tuple

vshlapakov: @chekunkov hm, that's true, but Dash doesn't support authorisation via `job_key:JWT_token`, are we going to add it for consistency? or you propose to leave it up to user, but put a list with jobkey-auth supported logic somewhere to the docs? if latter, should we check for it and print a warning when instantiating Connection?

chekunkov: > but Dash doesn't support authorisation via job_key:JWT_token, are we going to add it for consistency?

true, we already discuss this for one endpoint in Dash, we may expand it to a bigger scope, both in terms of supported endpoints and entities, e.g. we can authorize JWT tokens to operate on a project level.

but my original question was more about semantics. `apikey` scope is kinda narrow, while in reality we use other ways to authenticate clients, which are passed in the same way, so more general `auth` is more suitable in my opinion.

as for handling jwt tokens - I'd expect new client to be able to consume hex encoded jwt tokens as they are stored in environment and handle decoding and splitting it into (user, password) auth tuple internally

vshlapakov: I see, makes sense to me. Let's do what you're proposing 👌

vshlapakov: Added, please review [the changes](https://github.com/scrapinghub/python-scrapinghub/pull/48/commits/d5ef2c3ec84d3b8c48109efeb8e1c83b802e4ece).

Now it works in the following manner (I added a warning when using user/pass pair in legacy `Connection`): with a given `SHUB_JOBAUTH` env var, a client can be instantiated with:
```
c = ScrapinghubClient(os.env['SHUB_JOBAUTH']))
UserWarning: A lot of endpoints support authentication only via apikey.
```
After it you can use the client as usual, but it have access only to job data for now.

scrapinghub/utils.py
```diff
@@ -105,3 +107,28 @@ def format_iter_filters(params):
...
+    ('1/2/3', 'some.jwt.token')
+    """
+    if auth is None or isinstance(auth, tuple):
+        return auth
+    try:
+        auth = decode(auth, 'hex_codec')
+        if not isinstance(auth, string_types):
+            auth = auth.decode('ascii')
+    except binascii.Error:
+        pass
```
chekunkov: @vshlapakov did you try it with real apikey? it won't work correctly because real apikey consists of valid hexademical characters. you should try to decode the apikey, split it in parts, check that username part is a valid jobkey and password part is not empty - and only after that you can assume that this is a jwt token auth and can be used for auth. 

another option is to check apikey len, but this wouldn't allow us to change apikey len in future wihtout causing users to upgrade

scrapinghub/client.py
```diff
@@ -67,10 +68,15 @@ class ScrapinghubClient(object):
...
     """
 
-    def __init__(self, apikey=None, dash_endpoint=None, **kwargs):
+    def __init__(self, auth=None, dash_endpoint=None, **kwargs):
         self.projects = Projects(self)
-        self._connection = Connection(apikey=apikey, url=dash_endpoint)
-        self._hsclient = HubstorageClient(auth=apikey, **kwargs)
+        auth = parse_auth(auth)
+        connection_kwargs = {'apikey': auth, 'url': dash_endpoint}
+        if len(auth) == 2:
```
chekunkov: parse_auth should raise ValueError if auth tuple cannot be extracted or its len != 2. this way you'll always be able to unpack it into username/apikey and password part and pass them to target functions directly.

scrapinghub/utils.py
vshlapakov: You're right, I focused on jwt tokens yesterday and completely missed that apikey is also hex-encoded 🤦‍♂️  I tried to address the issues in the following commits, check when you have some time pls, corresponding tests are included.

scrapinghub/utils.py
```diff
@@ -112,23 +113,53 @@ def format_iter_filters(params):
...
+                               "environment variable not set")
+        return (apikey, '')
+
+    if isinstance(auth, tuple):
+        if len(auth) != 2:
+            raise ValueError("Wrong amount of authentication credentials")
+        if not isinstance(auth[0], string_types):
+            raise ValueError("Login must me of a string type")
+        if not (auth[1] is None or isinstance(auth[1], string_types)):
+            raise ValueError("Password must be None or of a string type")
```
chekunkov: let's keep it simple - both loging and password can be strings only, if password isn't set it's empty string

tests/client/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_parse_auth_bad_apikey():
+    with pytest.raises(ValueError):
+        parse_auth('apikey')
+
+
+def test_parse_auth_apikey():
+    test_key = u'\xe3\x98\xb8\xe6\x91\x84\xe9'
+    apikey = encode(test_key.encode('utf8'), 'hex_codec').decode('ascii')
```
chekunkov: wat?

tests/client/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+        parse_auth((1234, ''))
+
+
+def test_parse_auth_not_string():
+    with pytest.raises(ValueError):
+        parse_auth(12345)
+
+
+def test_parse_auth_simple():
+    assert parse_auth('user:pass') == ('user', 'pass')
```
chekunkov: you already have doctests in the parse_auth function, why not simply enable them?

scrapinghub/utils.py
```diff
@@ -83,3 +87,81 @@ def proxy_methods(origin, successor, methods):
...
+        raise ValueError("Wrong authentication credentials")
+
+    decoded_auth = None
+    try:
+        decoded_auth = decode(auth, 'hex_codec')
+    except (binascii.Error, TypeError):
+        login, _, password = auth.partition(':')
+        if not password:
+            raise ValueError("Bad apikey, please check your credentials")
+        return (login, password)
```
chekunkov: honestly, it's not super clear for me how you detect apikey vs jwt here? both apikey and jwt in my local tests pass this block without exception. can you split the process in 2 explicit parts - function that tries to extract jobkey and jwt from string, if it fails assume this is a apikey and validate it

tests/client/test_utils.py
vshlapakov: Right, will do

tests/client/test_utils.py
vshlapakov: I tried to show that apikey is just a utf8 string encoded with hex: all the conversions are related with that `codecs.encode` accepts a binary and returns a binary. But yeah, now I see it's not clear, let's just test with pre-generated test apikey

scrapinghub/utils.py
vshlapakov: > both apikey and jwt in my local tests pass this block without exception

That's right, both apikey and jwt are hex-encoded strings, so it should be decoded with `codecs.decode(auth, 'hex_codec')` on above. If there's `binascii.Error` or `TypeError`, we consider it as `user:password` pair, otherwise if there's no password - we assume that apikey wasn't correct.

Ok, let's try to split it to simplify the logic.

chekunkov: > both apikey and jwt are hex-encoded strings

@vshlapakov probably we shouldn't have such expectations for apikey, that may change in future.

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+addopts = --doctest-modules --doctest-glob='scrapinghub/*.py'
```
chekunkov: @vshlapakov you either need `--doctest-modules` OR  `--doctest-glob='scrapinghub/*.py'` (I think the latter)

scrapinghub/client.py
```diff
@@ -457,6 +457,13 @@ def schedule(self, spidername=None, **params):
             raise ValueError('Please provide spidername')
         params['project'] = self.projectid
         params['spider'] = spidername or self.spider.name
+        spider_args = params.pop('spider_args', None)
+        if spider_args:
+            if not isinstance(spider_args, dict):
+                raise ValueError("spider_args should be a dictionary")
+            cleaned_args = {k: v for k, v in spider_args.items()
+                            if k not in params}
+            params.update(cleaned_args)
```
chekunkov: @vshlapakov please update docs - spider_args should be the only recommended way to pass spider arguments

PR #49, opened 2017-02-28
Split frontier logic into Frontiers & Frontier

vshlapakov:
We also agreed to move frontier slot logic to a separate entity in a similar way.

Please, review.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -446,7 +446,8 @@ List all slots::
 
 Get a frontier slot by name::
 
-    >>> frontier.get('example.com')
+    >>> slot = frontier.get('example.com')
+    >>> slot
     <scrapinghub.client.FrontierSlot at 0x1049d8978>
 
 Add a request to the slot::
@@ -466,13 +467,17 @@ Add requests with additional parameters::
     >>> slot.q.add([{'fp': '/'}, {'fp': 'page1.html', 'p': 1, 'qdata': {'depth': 1}}])
     >>> slot.flush()
 
-To list requests for a given slot::
+To retrieve all requests for a given slot::
 
-    >>> reqs = slot.q.list()
+    >>> reqs = slot.q.iter()
 
-To retrieve fingerprints for a given slot::
+To retrieve all fingerprints for a given slot::
 
-    >>> fps = [req['requests'] for req in slot.q.iter()]
+    >>> fps = slot.f.iter()
+
+To list all the requests use ``list()`` method (similar for ``fingerprints``)::
+
+    >>> fps = slot.q.list()
 
 To delete a batch of requests::
 
--- a/README.rst
+++ b/README.rst
@@ -451,25 +451,32 @@ Get a frontier slot by name::
 
 Add a request to the slot::
 
-    >>> slot.add([{'fp': '/some/path.html'}])
+    >>> slot.queue.add([{'fp': '/some/path.html'}])
     >>> slot.flush()
+
+Add a fingerprint only to the slot::
+
+    >>> slot.fingerprints.add(['fp1', 'fp2'])
+    >>> slot.flush()
+
+There are convenient shortcuts: ``f`` for ``fingerprints`` and ``q`` for ``queue``.
 
 Add requests with additional parameters::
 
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -411,44 +411,74 @@ Usual workflow with `Collections`_ would be::
...
+
+Get a frontier by name::
+
+    >>> frontier = frontiers.get('test')
+    >>> frontier
+    <scrapinghub.client.Frontier at 0x1048ae4a8>
+
+Get an iterator to iterate through a frontier slots::
+
+    >>> frontier.iter_slots()
```
chekunkov: maybe replace with `iter()` and `list()`?

scrapinghub/client.py
```diff
@@ -1051,6 +1051,122 @@ def post(self, _value, **kwargs):
...
+        return self._frontiers._origin.delete(self.key, slot, ids)
+
+    def delete_slot(self, slot):
+        return self._frontiers._origin.delete_slot(self.key, slot)
+
+    def iter_slots(self):
+        return iter(self.list_slots())
+
+    def list_slots(self):
+        return next(self._frontiers._origin.apiget((self.key, 'list')))
```
chekunkov: Just a proposal - maybe we should add another entity - Slot or FrontierSlot? Frontier.iter() will go through available slot names, Frontier.get() will get Slot object, Slot.iter() or Slot.list() will do what read() does now, Slot.delete(ids) will delete data from the slot, Frontier.delete(name) will delete whole slot

vshlapakov: Makes sense, I'd go with FrontierSlot

README.rst
vshlapakov: I had doubts about it but if we're going to use a separate entity for slots, it looks fine

vshlapakov:
@chekunkov Ready for next review, I also added integration tests for frontier logic and added new cassettes to #40.

README.rst
```diff
@@ -411,44 +411,83 @@ Usual workflow with `Collections`_ would be::
...
 
-To retrieve fingerprints for a given slot::
+Flush data of the given frontier::
 
-    >>> fps = [req['requests'] for req in frontier.read('test', 'example.com')]
+    >>> frontier.flush()
+
+To delete the slot ``example.com`` from the frontier::
+
+    >>> frontier.delete('example.com')
```
chekunkov: just realized this can be confusing - usually it's expected to call delete on the object that should be deleted. e.g. we have Job.delete() method. What if we move `delete` to FrontierSlot, if no arguments are passed the entier slot is removed, if ids are passed - we remove only some requests?

vshlapakov: Agree, it's bit more consistent with other classes, let's do that

scrapinghub/client.py
```diff
@@ -1051,6 +1051,164 @@ def post(self, _value, **kwargs):
...
+        self.key = slot
+        self._client = client
+        self._frontier = frontier
+
+    def add(self, fps):
+        """Add requests to slot."""
+        origin = self._frontier._frontiers._origin
+        return origin.add(self._frontier.key, self.key, fps)
+
+    def iter(self, mincount=None):
```
chekunkov: better replace `mincount` with `**kwargs`, it's server responsibility to decide what parameters are valid or not

README.rst
```diff
@@ -451,25 +451,32 @@ Get a frontier slot by name::
...
 
 To list requests for a given slot::
 
-    >>> reqs = slot.list()
+    >>> reqs = slot.q.list()
 
 To retrieve fingerprints for a given slot::
 
-    >>> fps = [req['requests'] for req in slot.iter()]
+    >>> fps = [req['requests'] for req in slot.q.iter()]
```
chekunkov: mmm, shouldn't it be `slot.f.iter()` ?

chekunkov:
After we merge this PR we will probably need to split client.py into smaller modules - 1.3k lines is a bit too much

vshlapakov:
@chekunkov I think that's it, take a look pls when you have time.

PR #51, opened 2017-03-02
Frontier newcount counter per slot

vshlapakov:
Review, please.


tests/client/test_frontier.py
```diff
@@ -1,4 +1,4 @@
-
+import time
 from six import string_types
 from types import GeneratorType
 from collections import Iterable
```
chekunkov: could you please rearrange imports as PEP8 requires?

PR #52, opened 2017-03-02
Splitting client to several modules for simplicity

vshlapakov:
Review, please.


scrapinghub/client/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+    @wrap_http_errors
+    def _request(self, *args, **kwargs):
+        return super(Connection, self)._request(*args, **kwargs)
+
+
+class HubstorageClient(_HubstorageClient):
+
+    @wrap_http_errors
+    def request(self, *args, **kwargs):
+        return super(HubstorageClient, self).request(*args, **kwargs)
```
chekunkov: let's add `__all__ = ['ScrapinghubClient']` - to exclude possible clashes with HS client and Connection on import

scrapinghub/client/frontiers.py
```diff
@@ -0,0 +1,278 @@
+from functools import partial
+from collections import defaultdict
+
+from six import string_types
+
+from scrapinghub.hubstorage.frontier import Frontier as _Frontier
```
chekunkov: `..hubstorage.frontier` ?

PR #53, opened 2017-03-06
Minor improvements for new python-client

vshlapakov:
I created the additional PR to fix some minor issues before release:

- adding new client exceptions to top-level of the package
- proxying JobMeta & project Settings classes to get rid of old client methods and attributes
(these are last 2 classes used as-is in the new client, it'd be much more consistent to clean it too)
- sorting and cleaning tests to match latest changes in the client

Please, review.


scrapinghub/client/jobs.py
```diff
@@ -305,7 +306,7 @@ class Job(object):
         >>> job.metadata['state']
         'finished'
     """
-    def __init__(self, client, jobkey, metadata=None):
+    def __init__(self, client, jobkey):
```
vshlapakov: The constructor is not public, doesn't make sense to have a keyword arg

scrapinghub/client/utils.py
```diff
@@ -106,6 +107,20 @@ def __init__(self, cls, client, key):
             self._proxy_methods(methods)
             self._wrap_iter_methods([method[0] for method in methods])
 
+        if issubclass(cls, MappingResourceType):
+            methods = ['ignore_fields', 'expire', 'save', 'liveget',
+                       ('iter', '__iter__')]
+            self._proxy_methods(methods)
+            # the following methods are special and should be set to class
```
vshlapakov: To apply the special methods we have to set it on class level, and enforce rewriting it for `__str__`/`__repr__` - they always exist in any class

scrapinghub/client/utils.py
```diff
@@ -106,6 +107,20 @@ def __init__(self, cls, client, key):
             self._proxy_methods(methods)
             self._wrap_iter_methods([method[0] for method in methods])
 
+        if issubclass(cls, MappingResourceType):
+            methods = ['ignore_fields', 'expire', 'save', 'liveget',
```
vshlapakov: After some tests I think we shouldn't proxy `_cached` method here: looks like it was mainly used when instantiating JobMeta with some pre-cached data, but now JobMeta constructor is not public. We still could add it as a property to set it after instantiating jobmeta, but I'm not sure that it's actually used by anyone

scrapinghub/client/jobs.py
```diff
@@ -13,6 +13,7 @@
 from .samples import Samples
 from .exceptions import NotFound, InvalidUsage, DuplicateJobError
 from .utils import get_tags_for_update
+from .utils import _Proxy
 from .utils import parse_job_key
```
chekunkov: I'd merge ^ 3 lines

scrapinghub/client/jobs.py
```diff
@@ -442,3 +443,56 @@ def purge(self):
...
+
+        >>> job.metadata['my-meta'] = 'test'
+
+    - save updated metadata on server
+
+        >>> job.metadata.save()
+
+    - expire metadata local cache
+
+        >>> job.meta.expire()
```
chekunkov: .metadata?

scrapinghub/client/projects.py
```diff
@@ -121,4 +122,57 @@ def __init__(self, client, projectid):
...
+
+    - iterate through project settings keys
+
+        >>> project.settings.iter()
+        <dict_keyiterator at 0x106ae9cc8>
+
+    - list project settings keys
+
+        >>> project.settings.list()
+        ['default_job_units', 'job_runtime_limit']
```
chekunkov: actually not sure how much it makes sense to follow iter()/list() convention in case of Settings and Metadata because they are rather dict-like objects and there are expectations that they will behave as dicts (so keys(), iterkeys() should work instead of list(), iter())

scrapinghub/client/utils.py
chekunkov: Are you sure it's a good idea to mix this into the _Proxy? If we need a proxy for mapping resources - why cannot we create a separate class for that? + I'm not sure it's a good idea to follow the same conventions for dict-like objects we do for resources (see https://github.com/scrapinghub/python-scrapinghub/pull/53/files#r104448204)

vshlapakov: Not really, now it's clear that it's better to move the logic into a separate class like you proposed, and yeah - agree about the conventions issue, thanks!

scrapinghub/client/utils.py
```diff
@@ -140,6 +132,46 @@ def list(self, *args, **kwargs):
         return list(self.iter(*args, **kwargs))
 
 
+class _MappingProxy(_ProxyBase):
+
+    def __init__(self, cls, client, key):
+        super(_MappingProxy, self).__init__(cls, client, key)
+        methods = ['ignore_fields', 'expire', 'save', 'liveget']
+        self._proxy_methods(methods)
+        self.iterkeys = self.__iter__
```
chekunkov: I wonder, maybe we can ask _MappingProxy to map everything *except* some methods? I think it'll be a short list

scrapinghub/client/jobs.py
```diff
@@ -442,3 +441,56 @@ def purge(self):
...
+
+    - iterate through job metadata keys
+
+        >>> job.metadata.iter()
+        <dict_keyiterator at 0x1049ac188>
+
+    - list job metadata keys
+
+        >>> job.metadata.list()
+        ['project', 'tags', 'version']
```
chekunkov: don't forget to update docs

scrapinghub/client/utils.py
```diff
@@ -136,6 +183,8 @@ def wrapped(*args, **kwargs):
 def proxy_methods(origin, successor, methods):
     """A helper to proxy methods from origin to successor.
 
+    force param enforces rewriting attribute even if it exists in successor.
+
```
chekunkov: force param is gone already

vshlapakov:
@chekunkov I split `_Proxy` logic to a few classes with a recent commit, could you take a look pls? I wonder about some things:
- would it be better to move all the proxy logic into a separate module `proxy.py`
- do we need additional dict methods to the `MappingProxy` (like `update()`, `items()`, etc)

During refactoring I realised that it was a bad idea to proxy instance methods as class methods - it doesn't work as expected, and I replaced it with proper `__`-methods in `MappingProxy`. I also found a minor issue when trying to save a wrong setting:
```
>>> p.settings['xyz'] = 123
>>> p.settings.save()
InvalidUsage: Unknown project setting "xyz".
>>> del p.settings['xyz']
>>> p.settings.save()
InvalidUsage: no such settings "xyz"
```
So after unsuccessful attempt to save an unknown settings, it's not straightforward to fix it, the only option is `p.settings._origin._deleted.discard('xyz')` which is not obvious at all. Do we need to proxy `_deleted` attribute as well, or add some method to improve this?

scrapinghub/client/utils.py
vshlapakov: Not sure about pros, we can't proxy special methods with underscores and totally there're only 4 methods listed here above + `iterkeys`, I have doubts that blacklisting will simplify anything. But probably I didn't get the whole idea, could you explain pls?

chekunkov: Not sure why you can't. Use `dir(cls)` to iterate over attributes of `cls`, get all callables, filter methods should shouldn't stick in the new client, proxy everything else. Am I missing something?

scrapinghub/client/utils.py
```diff
@@ -136,40 +137,31 @@ class _MappingProxy(_ProxyBase):
...
+        blacklisted_attrs = ['__class__', '__init__', '__str__', '__repr__']
 
-    def keys(self):
-        return list(self.iterkeys())
+        # proxy all the main methods from MutableMapping
+        methods = [method for method in dir(MutableMapping)
+                   if callable(getattr(MutableMapping, method))
+                   and method not in origin_cls_methods
+                   and method not in blacklisted_attrs]
+        proxy_methods(MutableMapping, self.__class__, methods, force=True)
```
chekunkov: IMO you don't need to separate these methods from MutableMapping methods - as `self._origin` is an instance of a subclass of a MutableMapping you'll need only build a list of methods to proxy as 

```
mutable_mapping_methods = {method for method in dir(MutableMapping) if callable(getattr(MutableMapping, method))}
blacklisted_methods = {'__class__', '__init__', '__str__', '__repr__'}
resource_type_methods = {'ignore_fields', 'expire', '_data', '_deleted', 'save', 'liveget', ('iter', '__iter__')}
methods_to_proxy = (mutable_mapping_methods | resource_type_methods) - blacklisted_methods
```

and then proxy them **all** to the origin

```
proxy_methods(self._origin, self, methods_to_proxy)
```

scrapinghub/client/utils.py
```diff
@@ -136,40 +137,31 @@ class _MappingProxy(_ProxyBase):
...
+        proxy_methods(self._origin, self, other_attrs)
 
-    def get(self, key, default=None):
-        try:
-            return self[key]
-        except KeyError:
-            return default
+    @property
+    def _data(self):
+        return self._origin._data
```
chekunkov: You should probably remove it from `other_attrs` list?

scrapinghub/client/utils.py
```diff
@@ -143,7 +143,7 @@ def __init__(self, cls, client, key):
                               '__getitem__', '__setitem__', '__delitem__']
 
         other_attrs = ['ignore_fields', 'expire', '_data', '_deleted',
-                       'save', 'liveget', ('iter', '__iter__')]
+                       'save', 'liveget', ('iterkeys', '__iter__')]
```
chekunkov: note that MutableMapping has different set of methods in different Py versions

```
# alex in ~ [17:46:36]
→ python3.6
Python 3.6.0 (default, Feb 10 2017, 01:01:03)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import MutableMapping
>>> dir(MutableMapping)
['_MutableMapping__marker', '__abstractmethods__', '__class__', '__contains__', '__delattr__', '__delitem__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getitem__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__iter__', '__le__', '__len__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__reversed__', '__setattr__', '__setitem__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', '_abc_cache', '_abc_negative_cache', '_abc_negative_cache_version', '_abc_registry', 'clear', 'get', 'items', 'keys', 'pop', 'popitem', 'setdefault', 'update', 'values']
>>>

# alex in ~ [17:47:01]
→ python2.7
Python 2.7.10 (default, Jul 30 2016, 19:40:32)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import MutableMapping
>>> dir(MutableMapping)
['_MutableMapping__marker', '__abstractmethods__', '__class__', '__contains__', '__delattr__', '__delitem__', '__dict__', '__doc__', '__eq__', '__format__', '__getattribute__', '__getitem__', '__hash__', '__init__', '__iter__', '__len__', '__metaclass__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setitem__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_abc_cache', '_abc_negative_cache', '_abc_negative_cache_version', '_abc_registry', 'clear', 'get', 'items', 'iteritems', 'iterkeys', 'itervalues', 'keys', 'pop', 'popitem', 'setdefault', 'update', 'values']
>>>
```

There's no iterkeys method in Py 3

scrapinghub/client/utils.py
chekunkov: another thing to note - you can probably build a list of methods to blacklist like

```
[method for method in dir(self) if callable(getattr(self, method))]
```

vshlapakov: The proposed approach looks good, but as some special methods already exist in the object (like `__str__`, `__repr__` etc) - they should be proxied with `force` option (rewritten with origin method even if the method already exists in child class). If we proxy all the methods with `force` option - it means the methods cannot be redefined in subclasses. There's no need for this for now, but I wonder if we should reserve the option or it's a premature optimisation and doesn't worth efforts.

chekunkov: @vshlapakov I don't follow why special methods has to be proxied at the class level...

chekunkov: re magic methods that already exist - if you follow example I provided you'll receive a list of methods to proxy which looks like this for Py2:

```
set(['__metaclass__', 'popitem', '__getattribute__', 'pop', '__sizeof__', '__setattr__', '__reduce_ex__', '__new__', '__contains__', '__format__', '_deleted', 'itervalues', 'save', '__len__', '__ne__', '__getitem__', 'get', 'keys', '__reduce__', ('iter', '__iter__'), 'update', 'liveget', '__iter__', '__subclasshook__', 'iteritems', '_data', '__eq__', 'iterkeys', '__delitem__', 'setdefault', 'items', 'clear', '__setitem__', '__delattr__', 'ignore_fields', 'expire', 'values'])
```

does this list include methods that should be proxied with a force option?

chekunkov: and why do you need `force` option, cannot you exclude only attrs that are defined in the successor class? e.g. `not hasattr(successor.__class__, successor_name)` instead of `not hasattr(successor, successor_name)`?

vshlapakov: @chekunkov 
> I don't follow why special methods has to be proxied at the class level...
http://stackoverflow.com/questions/10484304/setting-special-methods-using-setattr
https://paste.scrapinghub.com/show/11632/
So in Python 3 it works via rewriting methods at the class level, and in Python 2 it doesn't work at all, what I'm missing here?

> does this list include methods that should be proxied with a force option?
It does, empty class object already contains a lot of the methods from that list:
```
In [9]: class X():
   ...:     pass
In [10]: dir(X())
Out[10]:
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__']
```

chekunkov: > So in Python 3 it works via rewriting methods at the class level, and in Python 2 it doesn't work at all, what I'm missing here?

ok, TIL ^. there's a Python recipe for proxying I've found http://code.activestate.com/recipes/496741-object-proxying/, is it of any help?

vshlapakov: Thanks for the link, it works like a full proxy but something should be added to make it more selective. The code becomes much more difficult and python-dependent, now I'm not so sure that it's a good idea to make JobMeta/Settings behave like real dict.

scrapinghub/client/utils.py
```diff
@@ -133,35 +125,33 @@ def list(self, *args, **kwargs):
...
-
-    def __init__(self, cls, client, key):
-        super(_MappingProxy, self).__init__(cls, client, key)
+class _MappingProxy(_Proxy):
 
-        # cls_methods methods will be imported from origin class
-        origin_cls_methods = ['__iter__', '__len__', '__str__', '__repr__',
-                              '__getitem__', '__setitem__', '__delitem__']
+    def count(self):
+        return len(next(self._origin.apiget()))
```
chekunkov: meh, it's not a real API call, I think users can simply do `len(obj.list())`

scrapinghub/client/utils.py
```diff
@@ -133,35 +125,33 @@ def list(self, *args, **kwargs):
...
-        # blacklisted attributes should be just skipped on proxying
-        blacklisted_attrs = ['__class__', '__init__', '__str__', '__repr__']
+    def set(self, key_or_values, value=None):
+        if isinstance(key_or_values, six.string_types):
+            self._origin.apipost(key_or_values,
+                                 data=json.dumps(value),
+                                 is_idempotent=True)
+        elif isinstance(key_or_values, dict):
+            data = next(self._origin.apiget())
+            data.update(key_or_values)
```
chekunkov: we should *set* the value, completely override it, and we should be explicit about that in the documentation. just a thin wrapper around API endpoint. No GET->update->POST routine.

scrapinghub/client/utils.py
vshlapakov: I added it only for consistency with other classes, `count` is almost everywhere. Dropping it?

scrapinghub/client/utils.py
vshlapakov: I'm not so sure about it: there're project settings and jobmeta that can't be overriden by users. It means that completely overriding doesn't make a lot of sense. Even if you have admin privileges - you have to be very careful about it to not delete some important settings by accident

scrapinghub/client/utils.py
chekunkov: Yes, let's drop it for now, I doubt it's useful

scrapinghub/client/utils.py
chekunkov: Project settings, cannot be overridden by users at all, they can be read and changed by staff only. That's how it works in JobQ now, maybe later we will provide read access and partial write access.

Maybe you're right, it can be dangerous... In this case the method should be named `update`, cause it updates metadata, not sets it all to the give value.

vshlapakov: > Project settings, cannot be overridden by users at all
Right, I messed it
> In this case the method that takes dict should be named `update`
Completely agree, let's rename it if you don't mind

vshlapakov: I have another proposal, what if we keep `set` method for setting a single value, and add `update` to do partial updates with a dict?

chekunkov: @vshlapakov that's exactly what I proposed above, sorry if it wasn't clear enough :)

vshlapakov: @chekunkov I think it's me who missed `the method that takes dict`, sorry :)

scrapinghub/client/utils.py
```diff
@@ -130,19 +130,19 @@ class _MappingProxy(_Proxy):
...
-            data = next(self._origin.apiget())
-            data.update(key_or_values)
-            self._origin.apipost(jl={k: v for k, v in six.iteritems(data)
-                                     if k not in self._origin.ignore_fields},
-                                 is_idempotent=True)
-        else:
-            raise InvalidUsage("key_or_values should be string or dict")
+    def set(self, key, value):
+        if not key or not isinstance(key, six.string_types):
+            raise InvalidUsage("key should be a string")
```
chekunkov: Either type of value error, because "scrapinghub.exceptions.InvalidUsage: Usually raised in case of 400 response from API."

To be honest not sure we need this check at all

vshlapakov:
The redundant check is dropped, also I split `test_proxy.py` to a few test modules to match latest changes in the client (now there're `test_requests.py`, `test_items.py` etc). All the tests are fixed, `sc1467-vcr` is rebased on top of that changes, Travis [build](https://travis-ci.org/scrapinghub/python-scrapinghub/builds/210483983).

scrapinghub/client/projects.py
```diff
@@ -121,4 +124,53 @@ def __init__(self, client, projectid):
...
+        ...                          'job_runtime_limit': 20})
+
+    - delete project setting by name
+
+        >>> project.settings.delete('job_runtime_limit')
+    """
+    def set(self, key, value):
+        # FIXME drop the method when post-by-key is implemented on server side
+        if not isinstance(key, six.string_types):
+            raise InvalidUsage("key should be a string")
```
chekunkov: again, kinda unexpected use of `InvalidUsage`, should be TypeError. What do you think about renaming `InvalidUsage` to `BadRequest` after the 400 response message? Just like it's done for Unauthorized and NotFound?

scrapinghub/client/utils.py
```diff
@@ -125,6 +125,30 @@ def list(self, *args, **kwargs):
...
+
+    def get(self, key):
+        return next(self._origin.apiget(key))
+
+    def set(self, key, value):
+        self._origin.apipost(key, data=json.dumps(value), is_idempotent=True)
+
+    def update(self, values):
+        if not isinstance(values, dict):
+            raise InvalidUsage("values should be a dict")
```
chekunkov: TypeError

scrapinghub/client/projects.py
vshlapakov: Yes, let's do that, I don't have objections

PR #54, opened 2017-03-14
Improve client to use it via IDE

vshlapakov:
The main goal of the PR are the following:
- provide return types everywhere where it could be useful,
- hint kwargs in methods to clarify which params can be passed,
- maybe work on style improvements also, i.e. params naming.

I'm also reviewing all the current docstrings and fixing any inconsistencies.

### DIFF
--- a/scrapinghub/client/__init__.py
+++ b/scrapinghub/client/__init__.py
@@ -51,44 +51,43 @@ class ScrapinghubClient(object):
                                       url=dash_endpoint)
         self._hsclient = HubstorageClient(auth=(login, password), **kwargs)
 
-    def get_project(self, projectid):
+    def get_project(self, project_id):
         """Get :class:`Project` instance with a given project id.
 
         The method is a shortcut for client.projects.get().
 
-        :param projectid: integer or string numeric project id.
+        :param project_id: integer or string numeric project id.
         :return: :class:`Project` object.
-        :rtype: scrapinghub.client.Project.
+        :rtype: scrapinghub.client.projects.Project
 
         Usage::
 
             >>> project = client.get_project(123)
             >>> project
-            <scrapinghub.client.Project at 0x106cdd6a0>
+            <scrapinghub.client.projects.Project at 0x106cdd6a0>
         """
-        return self.projects.get(parse_project_id(projectid))
+        return self.projects.get(parse_project_id(project_id))
 
-    def get_job(self, jobkey):
-        """Get Job with a given jobkey.
+    def get_job(self, job_key):
+        """Get Job with a given job key.
 
-        :param jobkey: job key string in format 'project/spider/job',
+        :param job_key: job key string in format 'project_id/spider_id/job_id',
             where all the components are integers.
         :return: :class:`Job` object.
-        :rtype: scrapinghub.client.Job.
+        :rtype: scrapinghub.client.jobs.Job
 
         Usage::
 
             >>> job = client.get_job('123/1/1')
             >>> job
-            <scrapinghub.client.Job at 0x10afe2eb1>
+            <scrapinghub.client.jobs.Job at 0x10afe2eb1>
         """
-        projectid = parse_job_key(jobkey).projectid
-        return self.projects.get(projectid).jobs.get(jobkey)
+        project_id = parse_job_key(job_key).project_id
+        return self.projects.get(project_id).jobs.get(job_key)
 
     def close(self, timeout=None):
         """Close client instance.
 
-        :param timeout: (optional) float timeout secs to stop everything
-            gracefully.
+        :param timeout: (optional) float timeout secs to stop gracefully.
         """
         self._hsclient.close(timeout=timeout)
--- a/scrapinghub/client/activity.py
+++ b/scrapinghub/client/activity.py
@@ -49,12 +49,17 @@ class Activity(_Proxy):
         self._wrap_iter_methods(['iter'])
 
     def add(self, values, **kwargs):
+        """Add new event to the project activity.
+
+        :param values: a single event or a list of events, where event is
+            represented with a dictionary of ('event', 'job', 'user') keys.
+        """
         if not isinstance(values, list):
             values = list(values)
         for activity in values:
             if not isinstance(activity, dict):
                 raise ValueError("Please pass events as dictionaries")
-            jobkey = activity.get('job')
-            if jobkey and parse_job_key(jobkey).projectid != self.key:
+            job_key = activity.get('job')
+            if job_key and parse_job_key(job_key).project_id != self.key:
                 raise ValueError('Please use same project id')
         self._origin.post(values, **kwargs)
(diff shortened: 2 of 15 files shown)


scrapinghub/client/activity.py
```diff
@@ -10,51 +10,51 @@ class Activity(_Proxy):
     Not a public constructor: use :class:`Project` instance to get a
     :class:`Activity` instance. See :attr:`Project.activity` attribute.
 
-    Please note that list() method can use a lot of memory and for a large
```
vshlapakov: The lines are moved to `_Proxy.list` method to avoid copy-pasting it everywhere.

vshlapakov:
@chekunkov Initial work is done, let me know if there's anything could be improved, I'm afraid that I has lost freshness of vision and can miss anything important. Most of the changes in the PR are related with indentation, renaming params and fixing docstrings.

scrapinghub/client/activity.py
chekunkov: The problem with moving this to a parent class is that it becomes invisible when you check help or docs for the method. E.g. I think we should build documentation based on docstrings later on, this line should be in each list method that can cause such issues.

scrapinghub/client/items.py
```diff
@@ -9,47 +9,44 @@ class Items(_Proxy):
...
+            }]
     """
 
     def _modify_iter_params(self, params):
         """Modify iter filter to convert offset to start parameter.
 
-        Returns:
-            dict: updated set of params
+        :return: a dict with updated set of params.
+        :rtype: dict.
```
chekunkov: please remove the dot

scrapinghub/client/jobs.py
```diff
@@ -22,30 +22,42 @@ class Jobs(object):
...
+            in millisecons.
+        :param \*\*params: (optional) other filter params.
 
-    def count(self, **params):
-        """Count jobs for a given set of parameters.
-
-        :param \*\*params: (optional) a set of filters to apply when counting
-            jobs (e.g. spider, state, has_tag, lacks_tag, startts and endts).
         :return: jobs count.
+        :rtype: int.
```
chekunkov: I don't think rtype value should have a trailing dot, please check across the repo

scrapinghub/client/jobs.py
```diff
@@ -55,96 +67,168 @@ def count(self, **params):
...
         """
+        # FIXME we double-check the params here, is there a better way?
+        # Simpler way would be to keep **params only here and point to iter(),
+        # but then we loose hinting kwargs for list() method.
+        filter_kwargs = [('spider', spider), ('state', state),
+                         ('count', count), ('start', start),
+                         ('has_tag', has_tag), ('lacks_tag', lacks_tag),
+                         ('startts', startts), ('endts', endts),
+                         ('meta', meta)]
+        params.update({k: v for k, v in filter_kwargs if v is not None})
```
chekunkov: possibly you can generalize filtering? something like

```
update_kwargs(kwargs, spider=spider, state=state, count=count,
               spidername=spider_name)
```

note the last one, you can use map to different name if necessary

scrapinghub/client/jobs.py
```diff
@@ -155,38 +239,40 @@ def schedule(self, spidername=None, **params):
...
+        if job_key.project_id != self.project_id:
             raise ValueError('Please use same project id')
-        if self.spider and jobkey.spiderid != self.spider._id:
+        if self.spider and job_key.spider_id != self.spider._id:
             raise ValueError('Please use same spider id')
-        return Job(self._client, str(jobkey))
+        return Job(self._client, str(job_key))
 
-    def summary(self, _queuename=None, **params):
+    # FIXME there must me spider_name instead of spider_id for consistency
```
chekunkov: `spider_name` or `spider`? list takes `spider`, schedule takes `spider_name` :) or maybe above methods should be filtered by `spider_id` for consistency with summary and iter_last?

scrapinghub/client/jobs.py
```diff
@@ -288,7 +382,7 @@ class Job(object):
     :class:`Jobs` instance to get a :class:`Job` instance. See
     :meth:`ScrapinghubClient.get_job` and :meth:`Jobs.get` methods.
 
-    :ivar projectid: in integer project id.
+    :ivar project_id: in integer project id.
```
chekunkov: "in" is not needed here

scrapinghub/client/jobs.py
```diff
@@ -422,35 +522,35 @@ class JobMeta(_MappingProxy):
 
     Usage::
```
chekunkov: I'm a bit confused, why Usage is followed by `::`? When rst renders this will produce a literal block, I don't think we want literal block for the whole usage description, I expect that it's a list of short descriptions followed by a literal block with example:


```
Usage:

- get job metadata instance::

    >>> job.metadata
    <scrapinghub.client.jobs.JobMeta at 0x10494f198>
```

scrapinghub/client/utils.py
```diff
@@ -160,11 +201,12 @@ def wrapped(*args, **kwargs):
...
-    - each string defines:
-        a successor method name to proxy 1:1 with origin method
-    - each tuple should consist of 2 strings:
-        a successor method name and an origin method name
+    Accepts a list with strings and tuples::
+
+        - each string defines:
+            a successor method name to proxy 1:1 with origin method
+        - each tuple should consist of 2 strings:
+            a successor method name and an origin method name
```
chekunkov: Why literal block `::`?

scrapinghub/client/jobs.py
vshlapakov: Sounds good, will do, thanks

scrapinghub/client/collections.py
```diff
@@ -111,35 +157,49 @@ def __init__(self, client, collections, coltype, colname):
             wrapped = wrap_kwargs(getattr(self, method), format_iter_filters)
             setattr(self, method, wrapped)
 
-    def list(self, *args, **kwargs):
+    def list(self, requests_params=None, **params):
```
vshlapakov: Oh, nice, I'll add it 👍

scrapinghub/client/__init__.py
```diff
@@ -43,52 +47,62 @@ class ScrapinghubClient(object):
         <scrapinghub.client.ScrapinghubClient at 0x1047af2e8>
     """
 
-    def __init__(self, auth=None, dash_endpoint=None, **kwargs):
+    # FIXME not sure it's worth to keep all HS kwargs here, most of them
+    # are appliable only to oldy Hubstorage client
```
vshlapakov: Well, let's keep only necessary kwargs then

scrapinghub/client/jobs.py
vshlapakov: Right, so `spider` or `spider_id` then..I'd vote for `spider`, name is easier to remember

scrapinghub/client/jobs.py
vshlapakov: You're right, I'll fix it everywhere 👌

vshlapakov:
Tests are [fixed](https://travis-ci.org/scrapinghub/python-scrapinghub/builds/212051570), merging.

PR #55, opened 2017-03-21
Handle ValueError from _decode_response

vshlapakov:
Please, review.

### DIFF
--- a/scrapinghub/client/exceptions.py
+++ b/scrapinghub/client/exceptions.py
@@ -52,6 +52,10 @@ class DuplicateJobError(ScrapinghubAPIError):
     pass
 
 
+class ServerError(ScrapinghubAPIError):
+    pass
+
+
 def wrap_http_errors(method):
     @wraps(method)
     def wrapped(*args, **kwargs):
@@ -69,6 +73,8 @@ def wrap_http_errors(method):
                 raise ValueTooLarge(http_error=exc)
             elif 400 <= status_code < 500:
                 raise ScrapinghubAPIError(http_error=exc)
+            elif 500 <= status_code < 600:
+                raise ServerError(http_error=exc)
             raise
         except APIError as exc:
             msg = exc.args[0]
@@ -80,6 +86,8 @@ def wrap_http_errors(method):
                 raise BadRequest(msg)
             elif exc._type == APIError.ERR_AUTH_ERROR:
                 raise Unauthorized(http_error=exc)
+            elif exc._type == APIError.ERR_SERVER_ERROR:
+                raise ServerError(http_error=exc)
             raise ScrapinghubAPIError(msg)
     return wrapped
 
--- a/scrapinghub/legacy.py
+++ b/scrapinghub/legacy.py
@@ -138,6 +138,11 @@ class Connection(object):
         return self._decode_response(response, format, raw)
 
     def _decode_response(self, response, format, raw):
+        if response.status_code == 404:
+            raise APIError("Not found", _type=APIError.ERR_NOT_FOUND)
+        elif 500 <= response.status_code < 600:
+            raise APIError("Internal server error",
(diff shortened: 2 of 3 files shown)


vshlapakov:
@chekunkov Please take a look and let me know if we need a better message for such cases, or more tests probably. I consider such issues as temporary ones, so the only advice is to retry the request later, maybe we also should create an additional subclass for `InternalServerError` representing HTTP 500 responses.

scrapinghub/client/exceptions.py
```diff
@@ -81,6 +81,11 @@ def wrapped(*args, **kwargs):
             elif exc._type == APIError.ERR_AUTH_ERROR:
                 raise Unauthorized(http_error=exc)
             raise ScrapinghubAPIError(msg)
+        except ValueError as exc:
+            # exception on attempt to decode wrong json in _decode_response
+            raise ScrapinghubAPIError(
+                "Error decoding server response ({}). Please try again later.".
+                format(str(exc)))
```
chekunkov: I don't think it's a good idea to catch all ValueErrors here and convert them to ScrapinghubAPIError. This is specifically a problem of legacy python-scrapinghub, we can fix it there, where it happens - e.g. by checking response code before trying to get json and raising and APIError. I Even if this will change behaviour - not a big deal, this is rather bug than a feature of legacy code. And I think that the new client desires major version bump, so we can allow ourselves to have some small incompatibilities in v2.0.0

vshlapakov: Ok, makes sense, let's handle it there 👍

scrapinghub/client/exceptions.py
```diff
@@ -69,6 +73,8 @@ def wrapped(*args, **kwargs):
                 raise ValueTooLarge(http_error=exc)
             elif 400 <= status_code < 500:
                 raise ScrapinghubAPIError(http_error=exc)
+            elif 500 <= status_code < 600:
+                raise ServerError(http_error=exc)
```
vshlapakov: @chekunkov Reworked a bit: the simplest way would be to call `raise_for_status` in the legacy client, and handle it here, does it look better?

chekunkov: @vshlapakov depending on whether it's possible to receive a non 200 response with some json body that should proceed to `self._decode_response` (not sure what's the return code for e.g. this error https://github.com/scrapinghub/python-scrapinghub/pull/55/files#diff-cd5ba6b088815a563525dde1900bbfecR150)

vshlapakov: That's a fair point, thanks. Return code is 403 for the authentication errors, so just calling `raise_for_status` doesn't fit. There's a few options how to handle it more gracefully in my mind:

- always raise HTTPError for all 500s errors(500<=status<600), w/o trying to parse body
- wrap `json.loads` with try-except and raise HTTPError from except clause (assuming that successful requests' responses always have proper body type, i.e. json)

My guess that 2nd approach is a way to go, and we also should do the same for jsonlines case. And as `ValueError` can happen in `_decode_response` on `json.loads` only, I'd wrap the whole method by catching `ValueError`s and raising `HTTPError`s based on response. Do you agree or I'm missing something?

chekunkov: The problem with the wrapper is that for jl format ValueError will never be catched as method returns generator in this case. I think it'd be reasonable to raise appropriate APIError for 404 (APIError.ERR_NOT_FOUND is never used) and 5xx (new err type) in _decode_response

P.S. let's change APIError error type values to strings vs current meaningless ints, e.g. `ERR_DEFAULT = 0` -> `ERR_DEFAULT = 'err_default'`

vshlapakov: Done, please review (but let me squash it before merge if it lgtm)

chekunkov:
LGTM, just squash it

PR #57, opened 2017-03-24
Improve documentation and prepare it for RTD

vshlapakov:
WIP


docs/conf.py
```diff
@@ -0,0 +1,171 @@
...
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'scrapinghub'
+copyright = u'2017, Pablo Hoffman, Daniel Graña'
```
chekunkov: I think it should be '2010-2017, Scrapinghub', and ending year should be generated from datetime.now().

wdyt @dangra?

docs/legacy/connection.rst
```diff
@@ -0,0 +1,108 @@
+scrapinghub.legacy.Connection
+=============================
+
+TODO add short description & deprecation message here!
```
chekunkov: don't forget to replace this TODO

chekunkov: or use http://www.sphinx-doc.org/en/stable/ext/todo.html#directive-todo for such notes

chekunkov: same for all following todos

scrapinghub/client/collections.py
```diff
@@ -91,7 +91,7 @@ def list(self):
 
         :return: a list of collections where each collection is
             represented by a dictionary with ('name','type') fields.
-        :rtype: list[dict]
+        :rtype: list of dicts
```
chekunkov: I think you should follow Sphinx examples for container types like [here](http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists) - which means keeping `list[dict]` like it was before.

... or typing notation https://docs.python.org/3/library/typing.html

docs/client/overview.rst
```diff
@@ -0,0 +1,598 @@
...
+
+    >>> job = project.jobs.get('123/1/2')
+    >>> job.key
+    '123/1/2'
+
+Also there's a shortcut to get same job with client instance::
+
+    >>> job = client.get_job('123/1/2')
+
+schedule
```
chekunkov: don't forget to rebase on top of master and update `schedule` -> `run`

docs/legacy/connection.rst
vshlapakov: Thanks, the TODOs are more for myself to not forget about it 👌

docs/client/overview.rst
vshlapakov: Sure, I will

scrapinghub/client/collections.py
vshlapakov: Initial approach led to a lot of Sphinx warnings like `more than one target found for cross-reference u'list'`, so using `list of dicts` was a workaround to overcome this. But Python typing notation (using `List`) works too, let's use it instead 👍

chekunkov: @vshlapakov try this fix https://github.com/tomchristie/django-rest-framework/issues/4882#issuecomment-277995050 (but it worth trying `:class:` prefix first - would be more appropriate)

scrapinghub/client/frontiers.py
```diff
@@ -105,7 +105,7 @@ def list(self):
         """List frontiers names.
 
         :return: a list of frontiers names.
-        :rtype: list[str]
+        :rtype: list of strings
```
chekunkov: missed this one

docs/client/overview.rst
```diff
@@ -0,0 +1,622 @@
...
+.. _spider:
+
+Spider
+------
+
+Like project instance, spider instance has ``jobs`` field to work with the spider's jobs.
+
+To schedule a spider run::
+
+    >>> spider.jobs.run(job_args={'arg1:'val1'})
```
chekunkov: missing quote char

docs/client/overview.rst
```diff
@@ -0,0 +1,622 @@
...
+Project
+-------
+
+Project instance has ``jobs`` field to work with the project jobs.
+
+Jobs instance is described well in ``Jobs`` section below.
+
+For example, to schedule a spider run (it returns a job object)::
+
+    >>> project.jobs.run('spider1', job_args={'arg1':'val1'})
```
chekunkov: missing space after `:`

docs/client/apidocs.rst
```diff
@@ -0,0 +1,110 @@
+Code-based documentation
```
chekunkov: I'd name this section `API Reference`, like http://flask.pocoo.org/docs/0.12/#api-reference

docs/client/apidocs.rst
```diff
@@ -0,0 +1,110 @@
...
+    :undoc-members:
+    :show-inheritance:
+
+frontiers
+---------
+
+.. automodule:: scrapinghub.client.frontiers
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
chekunkov: In frontiers documentation newcount is not documented

chekunkov:
There are a couple of public methods that are not documented, would be nice to fix that

code-based_documentation_ _scrapinghub_2_0_0_dev0_documentation

vshlapakov:
@chekunkov Thanks for the comments and sharp eye, updated!

scrapinghub/client/frontiers.py
```diff
@@ -184,6 +190,11 @@ def flush(self):
 
     @property
     def newcount(self):
+        """Amount of new entries added to frontier.
+
+        :return: amount of new entries
```
chekunkov: missing dots after `:return:` text

chekunkov:
Almost there! Can we also have links in Overview page to the API sections (e.g from https://github.com/scrapinghub/python-scrapinghub/pull/57/files#diff-c6024186d0575af110265c4516d13e3dR15 (projects) to https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/client/__init__.py#L36 or https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/client/projects.py#L15 ?

docs/client/overview.rst
```diff
@@ -0,0 +1,622 @@
...
+    '123'
+
+.. tip:: The above is a shortcut for ``client.projects.get(123)``.
+
+Project
+-------
+
+Project instance has ``jobs`` field to work with the project jobs.
+
+Jobs instance is described well in ``Jobs`` section below.
```
chekunkov: link?

docs/client/overview.rst
```diff
@@ -0,0 +1,622 @@
...
+scrapinghub.exceptions.BadRequest
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Usually raised in case of 400 response from API.
+
+
+scrapinghub.exceptions.Unauthorized
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Not enough access to some resources.
```
chekunkov: Confusing, better reword it like `Request lacks valid authentication credentials for the target resource`

chekunkov:
A couple more issues:

api_reference_ _scrapinghub_2_0_0_dev0_documentation

docs/client/apidocs.rst
```diff
@@ -0,0 +1,110 @@
...
+
+
+Submodules
+----------
+
+activity
+--------
```
chekunkov: should be under submodules section, use `~~~~~~~~~`

chekunkov: or, if you want a flat structure, follow an approach taken here http://flask.pocoo.org/docs/0.12/api/


```
API Reference
=============

Client object
-------------

Activity
--------

Collections
-----------

etc...
```

without splitting into `Module contents` and `Submodules`

docs/client/apidocs.rst
```diff
@@ -0,0 +1,110 @@
...
+
+.. automodule:: scrapinghub.client.activity
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
chekunkov: you need to add :inherited-members: everywhere because otherwise methods like `list` are not listed in the docs (http://stackoverflow.com/questions/30344289/how-can-i-show-inherited-members-of-a-class-in-my-sphinx-documentation)

api_reference_ _scrapinghub_2_0_0_dev0_documentation

chekunkov: also ideally `list()` on the above screenshot should reference proper list method in the Activity class

chekunkov:
Shouldn't these things be private and excluded from docs?

api_reference_ _scrapinghub_2_0_0_dev0_documentation

docs/client/apidocs.rst
```diff
@@ -0,0 +1,110 @@
...
+
+collections
+-----------
+
+.. automodule:: scrapinghub.client.collections
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
chekunkov: In case of collectoins it shows `Bases: scrapinghub.client.utils._Proxy`, not clickable, not documented - not useful.

docs/client/apidocs.rst
vshlapakov: There's a problem here, showing inherited members is fine, but it doesn't cover methods that we set in runtime when instantiating an object with `_proxy_methods` logic, i.e. `iter` [method](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/client/activity.py#L48) for `Activitity`.

scrapinghub/client/__init__.py
```diff
@@ -31,9 +31,10 @@ class ScrapinghubClient(object):
     :param auth: Scrapinghub APIKEY or other SH auth credentials.
     :param dash_endpoint: (optional) Scrapinghub Dash panel url.
     :param \*\*kwargs: (optional) Additional arguments for
-        :class:`scrapinghub.hubstorage.HubstorageClient` constructor.
+        :class:`~scrapinghub.hubstorage.HubstorageClient` constructor.
```
chekunkov: why?

scrapinghub/client/activity.py
```diff
@@ -7,11 +7,12 @@
...
 
-    Not a public constructor: use :class:`Project` instance to get a
-    :class:`Activity` instance. See :attr:`Project.activity` attribute.
+    Not a public constructor: use :class:`~scrapinghub.client.projects.Project`
+    instance to get a :class:`~scrapinghub.client.activity.Activity` instance.
+    See :attr:`~scrapinghub.client.projects.Project.activity` attribute.
 
-    Please note that list() method can use a lot of memory and for a large
-    amount of activities it's recommended to iterate through it via iter()
+    Please note that ``list()`` method can use a lot of memory and for a large
```
chekunkov: what about link to Activity.list here?

PR #61, opened 2017-03-28
Convert proxy logic to subclasses

vshlapakov:
As current proxy logic is based on adding a lot of methods in runtime on instantiating entities - all the methods are unavailable in documentation which is a serious drawback. It was decided to replace the approach with a set of subclasses of `_Proxy`, each is responsible for its own methods, and other missing methods should be added to target classes.

Some side changes in the PR:
- moving proxy logic to a separate module
- hiding `format_iter_filters` method
- dropping `iter_raw_*` methods
- minor documentation and docstrings fixes


scrapinghub/client/proxy.py
```diff
@@ -0,0 +1,202 @@
...
+        :param params: a dictionary with input parameters.
+        :return: an updated dictionary with parameters.
+        :rtype: :class:`dict`
+        """
+        return _format_iter_filters(params)
+
+
+class _ItemsResourceProxy(_Proxy):
+
+    def get(self, _key, **params):
```
chekunkov: missing docstring

TIL `items.get` :)

```
In [7]: job.items.get(0)
Out[7]:
{u'_type': u'Page',
 u'referer': None,
 u'size': u'24836',
 u'title': u'Web Crawling Platform & Data as a Service | Scrapinghub',
 u'url': u'https://scrapinghub.com/'}
```

PR #65, opened 2017-03-29
Fix API docs link

redapple:
https://doc.scrapinghub.com/api.html yields a 404


redapple:
There are a few other places with old links.

vshlapakov:
Thanks @redapple!

PR #69, opened 2017-04-02
Added: `scrapinghub.hustorage.frontier.Frontier.count_slot`

starrify:
This is a copy of scrapinghub/python-hubstorage#76


vshlapakov:
Hey @starrify, sorry for the late reply, is it still actual? I believe we don't want to modify `scrapinghub.hubstorage` client as it's outdated, but I could probably add similar functionality into updated `ScrapinghubClient`. Would be also nice to avoid traversing all the records in slot just to get count number, maybe we could add its support to our backend, but for now I'm not sure how much time it will take.

Gallaecio:
Closing in favor of https://github.com/scrapinghub/python-scrapinghub/pull/155.

PR #72, opened 2017-04-17
Add VCR.py json-serialized tests

vshlapakov:
The PR adds additional tests formed with json serialization based on:
- new vcrpy accept-header matcher to distinguish messagepack-marked requests from others
- additional fixture `json_and_msgpack` to parametrize some of the existing tests
- json-serialized tests being put to separate vcrpy cassettes, named with `-json.gz` suffix
- pytest `--disable-msgpack` option to disable msgpack tests (same if msgpack is unavailable)

I have some doubts if we need to add Travis configuration to run all the tests with uninstalled msgpack to make sure everything works as expected w/o msgpack at all, let's discuss it if needed.

Please, review.

PR #73, opened 2017-04-19
Add client parameter to disable msgpack use

vshlapakov:
Review, please.

The changes is a bit overlapped with https://github.com/scrapinghub/python-scrapinghub/pull/72 but not too much, anyway I can merge it into that PR if needed. If the approach looks good, I'll also add some tests.

### DIFF
--- a/scrapinghub/hubstorage/collectionsrt.py
+++ b/scrapinghub/hubstorage/collectionsrt.py
@@ -32,7 +32,7 @@ class Collections(DownloadableResource):
         :param path: None, tuple or string
 
         """
-        if not MSGPACK_AVAILABLE:
+        if not MSGPACK_AVAILABLE or not self.client.use_msgpack:
             return False
         path = urlpathjoin(path or '')
         match = COLLECTIONS_MSGPACK_REGEX.match(path)
--- a/scrapinghub/hubstorage/collectionsrt.py
+++ b/scrapinghub/hubstorage/collectionsrt.py
@@ -32,7 +32,7 @@ class Collections(DownloadableResource):
         :param path: None, tuple or string
 
         """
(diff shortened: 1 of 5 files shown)


scrapinghub/hubstorage/collectionsrt.py
```diff
@@ -32,7 +32,7 @@ def _allows_mpack(self, path=None):
         :param path: None, tuple or string
 
         """
-        if not MSGPACK_AVAILABLE:
+        if not MSGPACK_AVAILABLE or not self.client.use_msgpack:
```
chekunkov: quick question, why not set `self.use_msgpack = MSGPACK_AVAILABLE and use_msgpack` in init and check `if not self.client.use_msgpack` only?

vshlapakov: Looks better, thanks. Btw in both cases if messagepack is not available - it will silently ignore `use_msgpack` option, does it make sense to add some warnings for such case?

chekunkov: yes, good idea

vshlapakov: Added a warning message to illustrate what I meant in https://github.com/scrapinghub/python-scrapinghub/pull/73/commits/736e57da91833a3ceb9b6e455c61fa4e795c85d1.

scrapinghub/hubstorage/resourcetype.py
```diff
@@ -32,7 +32,7 @@ def _allows_mpack(self, path=None):
         :param path: None, tuple or string
 
         """
-        if not MSGPACK_AVAILABLE:
```
chekunkov: can you also drop unused imports?

Issue #79, opened 2017-08-17
'Items' object is not callable


ijharulislam:
I am using python scrapinghub. Now I am facing two problems. 
1. I am getting en error.

```
Traceback (most recent call last):
  File "shop_info/test.py", line 58, in <module>
    for item in job.items():
TypeError: 'Items' object is not callable
```

2. I want to pass a JSON object as an argument. I tried with this.
project.jobs.run('shop_info', job_args={'input_data': input_data})

input_data is a json object.

spookyUnknownUser:
I think this is supposed to be 
`job.items.iter()`

According to the pydoc
```
    - retrieve all scraped items from a job::

        >>> job.items.iter()
        <generator object mpdecode at 0x10f5f3aa0>
```

Seems like this wasn't updated in the quickstart.

ijharulislam:
Thank you for answer. I already fixed the issue with items.iter().

vshlapakov:
Thanks :+1:

PR #81, opened 2017-09-09
Update quickstart with correct items call

spookyUnknownUser:
See #79

Saying `job.items()` results in a callable error, I think the right syntax is the one in the pydoc: `job.items.iter()`

Right❓


vshlapakov:
Indeed, thanks and sorry for delay, have just saw the issue.

PR #85, opened 2017-10-30
frontier: collect newcounts per frontier/slot

immerrr:
This would be great to collect statistics about incremental crawl coverage.

@dangra @vshlapakov wdyt?


immerrr:
While patching tests I've realized that there's already a version of HSFrontier with newcounts per slot in `client.frontiers`. I wonder what's the best way to merge them...

immerrr:
I'm closing this PR as there's already such functionality provided in the new client.

PR #86, opened 2017-10-31
Improve tests

immerrr:
This PR adds pytest-catchlog for reviewing errors sent to logs, e.g. from background threads.

It also adds a rather unique frontier name for each test case so that test case ordering doesn't affect the vcrpy cassettes.


immerrr:
Sweet, I like them, merged your changes into my branch right away :)

tests/hubstorage/test_frontier.py
```diff
@@ -3,126 +3,125 @@
...
     fps3 = [{'fp': '/index_%s.html' % fp}
             for fp in range(batch_size * 2, batch_size * 3)]
-    frontier.add(TEST_FRONTIER_NAME, TEST_FRONTIER_SLOT, fps3)
+    frontier.add(frontier_name, TEST_FRONTIER_SLOT, fps3)
     frontier.flush()
 
     assert frontier.newcount == 150 + old_count
 
     # insert repeated fingerprints
-    fps4 = [{'fp': '/index_%s.html' % fp} for fp in range(0, batch_size)]
```
vshlapakov: @immerrr last question about the PR before merge, why the test with repeated fps was changed?

immerrr: Because `fps4` was unused and `fps3` were used as "repeated fingerprints" instead.

vshlapakov: Oh, thanks for cleaning it, I didn't spot it even now :man_facepalming:

vshlapakov:
Thanks again @immerrr! :+1:

PR #87, opened 2017-11-01
Minor changes and cassettes for #86

vshlapakov:
Proposals to improve https://github.com/scrapinghub/python-scrapinghub/pull/86.

1) I think it's better to stick with simple names, like `frontier_name`, as a frontier is a pretty much concrete logic entity within all the repository (and it's aligned with updated naming in the tests for official client).
2) Also reworked `test_frontier_newcount` a bit, as in contrast to other frontier-related tests it checks for global `newcount` counter for a project and it requires proper cleaning (added a small helper).
3) And replaces test cassettes, you can't prepare it without all the infrastructure running locally.

Please take a look.


immerrr:
LGTM

PR #88, opened 2017-11-14
Send request data as str/bytes

vshlapakov:
A short fix to overcome https://github.com/shazow/urllib3/issues/717 related to large POST requests via SSL raising `SSL3_WRITE_PENDING` error.

### DIFF
--- a/scrapinghub/hubstorage/resourcetype.py
+++ b/scrapinghub/hubstorage/resourcetype.py
@@ -54,7 +54,9 @@ class ResourceType(object):
         kwargs['url'] = urlpathjoin(self.url, _path)
         kwargs.setdefault('auth', self.auth)
         if 'jl' in kwargs:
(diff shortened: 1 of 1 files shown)


scrapinghub/hubstorage/resourcetype.py
```diff
@@ -54,7 +54,7 @@ def _iter_lines(self, _path, **kwargs):
         kwargs['url'] = urlpathjoin(self.url, _path)
         kwargs.setdefault('auth', self.auth)
         if 'jl' in kwargs:
-            kwargs['data'] = jlencode(kwargs.pop('jl'))
```
dangra: I think we need a comment here explaining why the `.encode()` is needed and what upstream bug it is preventing.

vshlapakov: @dangra Good point, updated :ok_hand:

Issue #19, opened 2015-02-13
No default timeout for HTTP calls


redapple:
Looks like python-requests uses `timeout=None` by default.
Calls to the API can hang a long time (forever?)

It would be helpful to add a [timeouts](http://docs.python-requests.org/en/latest/user/advanced/?highlight=timeout#timeouts) for GET and POST calls

dangra:
:+1:

dangra:
fixed by #94

PR #94, opened 2018-01-15
Provide unified connection timeout

vshlapakov:
Addressing #19.

`HubstorageClient` already relies on `connection_timeout` param, I believe we want to keep its current behaviour and reuse it for `Connection` (with default to `None`, again - to keep the current logic).

### DIFF
--- a/scrapinghub/client/__init__.py
+++ b/scrapinghub/client/__init__.py
@@ -49,13 +49,13 @@ class ScrapinghubClient(object):
                  connection_timeout=DEFAULT_CONNECTION_TIMEOUT, **kwargs):
         self.projects = Projects(self)
         login, password = parse_auth(auth)
+        timeout = connection_timeout or DEFAULT_CONNECTION_TIMEOUT
         self._connection = Connection(apikey=login,
                                       password=password,
                                       url=dash_endpoint,
-                                      connection_timeout=connection_timeout)
+                                      connection_timeout=timeout)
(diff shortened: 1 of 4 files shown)


dangra:
As far as I understand current lack of default timeout is a bug in `legacy.Connection`, but not the case for [HubstorageClient  which defaults to 60s](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/hubstorage/client.py#L82). I think we should make the default timeout (60s) part of `ScrapinghubClient` too.

scrapinghub/client/__init__.py
```diff
@@ -43,13 +45,17 @@ class ScrapinghubClient(object):
         <scrapinghub.client.ScrapinghubClient at 0x1047af2e8>
     """
 
-    def __init__(self, auth=None, dash_endpoint=None, **kwargs):
+    def __init__(self, auth=None, dash_endpoint=None,
+                 connection_timeout=DEFAULT_CONNECTION_TIMEOUT, **kwargs):
```
dangra: `HubstorageClient` always sets a connection timeout even if you pass `None` or `0`, I think we should do the same as https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/hubstorage/client.py#L82

vshlapakov: Good point, I did it in this way initially but then decided to "optimize" the changes, thanks 🙂

PR #96, opened 2018-01-22
Fix parse_job_key issue on Python 3

jesuslosada:
`map` uses lazy evaluation on Python 3, so job_key values that contain strings that can't be parsed as integers don't trigger the ValueError exception.


vshlapakov:
Hey @jesuslosada, thanks for the patch 👍 Could you please add some basic test as well?

jesuslosada:
@vshlapakov Done

vshlapakov:
Perfect, thanks @jesuslosada!

PR #98, opened 2018-06-04
Docs: copy/paste typo

stav:


vshlapakov:
Thanks a lot @stav! :+1: I'll fix the failed tests a bit later.

stav:
What is the cause of the failed tests?

vshlapakov:
```
wheel requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*' but the running Python is 3.3.6
```
Lools like it comes from https://github.com/pypa/wheel/pull/216/files. We probably just need to consider Python 3.3 as unmaintained and drop the environment from our Travis tests.

Issue #103, opened 2018-10-24
Use SHUB_JOBAUTH environment variable in utils.parse_auth method


victor-torres:
Currently, the parse_auth method tries to get the API Key from the SH_APIKEY environment variable, which needs to be manually set either in spider's code or in the [docker] image's code. A regular practice is to create dummy users and associate them with the project so that real contributors don't have to share their API Keys.

Another option is to use the credentials provided by the SHUB_JOBAUTH, defined during runtime when executing jobs in the Scrapy Cloud platform. 

Although it's possible to use Collections and Frontera, this is not a regular Dash API Key but a JWT Token generated in runtime by JobQ service, which works only for a part of our API endpoints (JobQ/Hubstorage).

I'd like to contribute with a Pull Request adding support for this ephemeral API Key.

kalessin:
I personally like that feature. In the past that tocken couldn't be used to access frontier or collections, but seems it is possible now.

dangra:
> I personally like that feature. In the past that token couldn't be used to access frontier or collections, but seems it is possible now.

Not sure when this happened though. I have no objections as long as the token works for all api endpoints where it is used if SHUB_JOBAUTH is in the environment.

victor-torres:
I cannot guarantee if it will work for every endpoint provided by this library, but it seems reasonable to prioritize the previous authentication methods and use SHUB_JOBAUTH as a fallback when running in cloud environments. What do you think?

kalessin:
@dangra If the token doesn't work for some endpoint, nothing changes for the user that need to use it: it will just get an authentication error because no explicit apikey was passed. Notice that the feature is to use SHUB_JOBAUTH only if no explicit apikey was passed.

vshlapakov:
It won't work for all API endpoints, that's for sure. But I like the idea of the fallback: if `SH_APIKEY` is not provided, check for `SHUB_JOBAUTH` and if there's a valid JWT token - use it with a warning describing its limitations and what should be done to overcome it.

victor-torres:
I have just submitted a Pull Request with that suggestion (#104)

vshlapakov:
Merged, thanks!

PR #104, opened 2018-10-25
fallback to SHUB_JOBAUTH environment variable if SH_APIKEY is not set

victor-torres:
you can test with the command `py.test tests/client/test_utils.py`

I'm also adding requirements-base.txt to the requirements-test.txt file as we need some libs installed from this file to perform some tests

This fixes issue #103

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,5 +1,3 @@
--r requirements-base.txt
-
 mock
 vcrpy==1.10.3
 # FIXME remove the constraint after resolving
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,3 +1,5 @@
+-r requirements-base.txt
+
 mock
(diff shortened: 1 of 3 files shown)


requirements-test.txt
```diff
@@ -1,3 +1,5 @@
+-r requirements-base.txt
```
vshlapakov: I think the general idea here is:
- when you need to simply install a package, you install everything from requirements.txt
- when you want to install test-related packages, you install it **additionally**, with separate `pip install -r` cmd

So for me it's a matter of taste and I don't see any strong reason to change it.
Do you agree or I'm missing anything else here?

victor-torres:
Great. I have removed the changes from the requirements file in order to keep the scope of this PR attached to the original discussion.

vshlapakov:
Thanks @victor-torres!

PR #108, opened 2019-01-18
Docs about auth being optional in ScrapinghubClient

ejulio:
It wasn't documented that `auth` is optional and it reads from environment variables available in scrapy cloud.

### DIFF
--- a/scrapinghub/client/__init__.py
+++ b/scrapinghub/client/__init__.py
@@ -32,23 +32,11 @@ class ScrapinghubClient(object):
     :param auth: (optional) Scrapinghub APIKEY or other SH auth credentials.
         If not provided, it will read, respectively, from 
         ``SH_APIKEY`` or ``SHUB_JOBAUTH`` environment variables.
-        ``SH_APIKEY`` must be set when deploying the spider and
-        ``SHUB_JOBAUTH`` is an internal environment variable
-        available in *Scrapy Cloud* and can't be used locally.
-        Also note that some functionality might not be available through
-        ``SHUB_JOBAUTH``, but most of the Hubstorage public endpoints
-        can be accessed.
-        The following endpoints are available:
-
-        * job data endpoints (``/items/``, ``/logs/``, ``/requests/``, ``/samples/``)
-        * job metadata endpoints (``/meta/``)
-        * project collection endpoints (``/collections/``)
-        * project activity endpoints (``/activity/``)
-        * hub crawl frontier endpoints (``/hcf/``)
-
-        Please refer to the 
-        `oficial docs <https://doc.scrapinghub.com/scrapy-cloud.html>`_
-        for more information about the endpoints.
+        ``SHUB_JOBAUTH`` is available by default in *Scrapy Cloud*, but it does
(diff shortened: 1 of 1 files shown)


scrapinghub/client/__init__.py
```diff
@@ -29,7 +29,9 @@ def request(self, *args, **kwargs):
 class ScrapinghubClient(object):
     """Main class to work with Scrapinghub API.
 
-    :param auth: Scrapinghub APIKEY or other SH auth credentials.
+    :param auth: (optional) Scrapinghub APIKEY or other SH auth credentials.
+        If not provided, it will read, respectively, from 
+        ``SH_APIKEY`` or ``SHUB_JOBAUTH`` environment variables.
```
victor-torres: You could add two things here:

- the SHUB_JOBAUTH is available when executing on Scrapy Cloud
- the token contained in SHUB_JOBAUTH may not be authorized to access some endpoints

It's not permitted to schedule jobs using SHUB_JOBAUTH, for example. But I believe it's possible to read/write/remove Collections and Frontiers, for example.

vshlapakov: Good point, let's clarify it to avoid any confusion :+1:

ejulio: :+1: 
Is there a link/documentation mentioning which services will work with this JWT Token?
I've seen something in this issue (https://github.com/scrapinghub/python-scrapinghub/issues/103) from @victor-torres . But I could not find anything in scrapy cloud docs :(

ejulio: I just updated the docs with the info above.
Can you take another look? :smile:

scrapinghub/client/__init__.py
```diff
@@ -32,6 +32,23 @@ class ScrapinghubClient(object):
...
+        The following endpoints are available:
+
+        * job data endpoints (``/items/``, ``/logs/``, ``/requests/``, ``/samples/``)
+        * job metadata endpoints (``/meta/``)
+        * project collection endpoints (``/collections/``)
+        * project activity endpoints (``/activity/``)
+        * hub crawl frontier endpoints (``/hcf/``)
+
+        Please refer to the 
+        `oficial docs <https://doc.scrapinghub.com/scrapy-cloud.html>`_
```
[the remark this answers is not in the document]
ejulio: I think this is because it is an inline link http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks

scrapinghub/client/__init__.py
```diff
@@ -32,6 +32,23 @@ class ScrapinghubClient(object):
     :param auth: (optional) Scrapinghub APIKEY or other SH auth credentials.
         If not provided, it will read, respectively, from 
         ``SH_APIKEY`` or ``SHUB_JOBAUTH`` environment variables.
+        ``SH_APIKEY`` must be set when deploying the spider and
```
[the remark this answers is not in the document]
ejulio: hey @victor-torres , I just change to address your ideas.
Can you take another look?
Thanks!

vshlapakov:
Thanks guys @ejulio @victor-torres :+1:

PR #109, opened 2019-02-07
Add long_description to setup.cfg

manycoding:
https://pypi.org/project/scrapinghub/ doesn't contain any description

The result of running twine check after the changes:
```
(python-scrapinghub) bash-3.2$ twine check dist/*
Checking distribution dist/scrapinghub-2.1.0-py2.py3-none-any.whl: Passed
Checking distribution dist/scrapinghub-2.1.0.tar.gz: Passed
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,6 @@
+[metadata]
+long_description = file: README.rst
+long_description_content_type = text/x-rst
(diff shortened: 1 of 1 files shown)


manycoding:
@vshlapakov

PR #110, opened 2019-02-07
Unpin pytest

manycoding:
Just a small update.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,8 +1,6 @@
 mock
 vcrpy==1.10.3
-# FIXME remove the constraint after resolving
-# https://github.com/pytest-dev/pytest/issues/2966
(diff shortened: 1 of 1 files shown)


manycoding:
Wow, I am not sure why it fails.

vshlapakov:
I did some tests, it seems there're breaking changes b/w pytest 3.2.5, that we use now, and the recent one 4.3.3, tests start failing on 3.7.1, that's reproducible locally. I do want to upgrade all the test dependencies (incl vcrpy) and simplify the whole cassettes-based testing approach, that's something for the next version of the library.

manycoding:
@vshlapakov I created #120 to track it and am closing this one.

vshlapakov:
Sounds good, thanks.

PR #111, opened 2019-02-11
Add truncate method to collections

victor-torres:
This makes possible to delete an entire Collection with a single API request, without having to iterate through records and therefore making multiple API requests.


victor-torres:
@vshlapakov @raphapassini @ejulio 

Could you review this pull request, please?

raphapassini:
The code looks nice, I think you need to document this as well :)

victor-torres:
> The code looks nice, I think you need to document this as well :)

Thanks. I've just updated the source code improving documentation, @raphapassini.

vshlapakov:
@ejulio Agree, but the warning is related with a whole bunch of operations and not directly related with the changes, let's do it separately.

PR #112, opened 2019-02-12
Refactoring VcrGzipSerializer to normalize new cassettes

victor-torres:
With these changes, it would be easier to add new cassettes when adding new tests and making calls to HubStorage and Dash APIs.

For example, you can use a dummy Scrapy Cloud project and its deploy API Key to add a new test cassette as following:

```
HS_PROJECT_ID=<project id> HS_AUTH=<api key> AUTH=<api key> USER_AUTH=<api key> HS_ENDPOINT= DASH_ENDPOINT= py.test tests/client/test_collections.py::test_new_case
```

This will use the production servers with the provided credentials and will create a new cassette file at `tests/client/cassettes/test_collections/test_new_case.gz`.

From now on you can execute the test with just:

```
py.test tests/client/test_collections.py::test_new_case
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,16 +47,7 @@ class VCRGzipSerializer(object):
 
 
 def normalize_endpoint(uri, endpoint, default_endpoint):
-    old = endpoint
-    new = default_endpoint
-
-    if old.endswith('/'):
-        old = old[:-1]
-
-    if new.endswith('/'):
-        new = new[:-1]
-
-    return uri.replace(old, new)
+    return uri.replace(endpoint.rstrip('/'), default_endpoint.rstrip('/'))
 
 
 def normalize_cassette(cassette_dict):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,99 @@
 # -*- coding: utf-8 -*-
+import base64
+import os
+import pickle
 import pytest
 import re
+import zlib
 
 from scrapinghub.hubstorage.serialization import MSGPACK_AVAILABLE
+from scrapinghub import HubstorageClient
+from scrapinghub.legacy import Connection
+
+
+DEFAULT_PROJECT_ID = "2222222"
+DEFAULT_ENDPOINT = 'http://storage.vm.scrapinghub.com'
+DEFAULT_DASH_ENDPOINT = 'http://33.33.33.51:8080/api/'
+
+
+TEST_PROJECT_ID = os.getenv('HS_PROJECT_ID', DEFAULT_PROJECT_ID)
+TEST_SPIDER_NAME = 'hs-test-spider'
+TEST_FRONTIER_SLOT = 'site.com'
+TEST_BOTGROUP = 'python-hubstorage-test'
+TEST_COLLECTION_NAME = "test_collection_123"
+TEST_AUTH = os.getenv('HS_AUTH', 'f' * 32)
+TEST_ENDPOINT = os.getenv('HS_ENDPOINT', DEFAULT_ENDPOINT)
+TEST_COLLECTION_NAME = "test_collection_123"
+TEST_ADMIN_AUTH = os.getenv('AUTH', 'f' * 32)
+TEST_USER_AUTH = os.getenv('USER_AUTH', 'e' * 32)
+TEST_DASH_ENDPOINT = os.getenv('DASH_ENDPOINT', DEFAULT_DASH_ENDPOINT)
+
+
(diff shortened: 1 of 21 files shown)


tests/conftest.py
```diff
@@ -1,8 +1,99 @@
...
+        return base64.b64encode(compressed).decode('utf8')
+
+    def deserialize(self, cassette_string):
+        # receives a string, must return a dict
+        decoded = base64.b64decode(cassette_string.encode('utf8'))
+        return pickle.loads(zlib.decompress(decoded))
+
+
+def normalize_endpoint(uri, endpoint, default_endpoint):
+    old = endpoint
```
vshlapakov: We could replace the function with one-liner, wdyt?
```
def normalize_endpoint(uri, endpoint, default_endpoint): 
    return uri.replace(endpoint.rstrip('/'), default_endpoint.rstrip('/'))
```

tests/conftest.py
```diff
@@ -1,8 +1,99 @@
...
+        uri = normalize_endpoint(uri, dash_endpoint, DEFAULT_DASH_ENDPOINT)
+
+        interaction['request']['uri'] = uri
+
+        if 'Authorization' in interaction['request']['headers']:
+            del interaction['request']['headers']['Authorization']
+            interaction['request']['headers']['Authorization'] = (
+                'Basic {}'.format(
+                    base64.b64encode(
+                        '{}:'.format('f' * 32).encode('utf-8')
```
vshlapakov: That's exactly `TEST_ADMIN_AUTH`, isn't it?

tests/conftest.py
```diff
@@ -1,8 +1,99 @@
...
+        hs_endpoint = TEST_ENDPOINT or HubstorageClient.DEFAULT_ENDPOINT
+        uri = normalize_endpoint(uri, hs_endpoint, DEFAULT_ENDPOINT)
+
+        dash_endpoint = TEST_DASH_ENDPOINT or Connection.DEFAULT_ENDPOINT
+        uri = normalize_endpoint(uri, dash_endpoint, DEFAULT_DASH_ENDPOINT)
+
+        interaction['request']['uri'] = uri
+
+        if 'Authorization' in interaction['request']['headers']:
+            del interaction['request']['headers']['Authorization']
```
vshlapakov: Am I right that all cassette requests are stored as if they're all done with admin API key? I'm not sure it's correct as some requests explicitly use a regular user' `TEST_USER_AUTH` key to test for correct permissions handling (well, not many of them but that was the initial intention to use 2 different API keys in tests, with different access level). Is the unification really needed?

victor-torres:
Thank you for your comments here, @vshlapakov. I'm just waiting for some time before jumping into this PR again.

tests/conftest.py
victor-torres: My intention here was to obfuscate the API Keys used to perform the calls when storing the payload on cassettes.

I might be wrong, but I believe different API Keys will indeed be used when making the real API call, they'll just be replaced by a placeholder when being stored to avoid credentials leakage.

tests/conftest.py
victor-torres: Sure.

tests/conftest.py
victor-torres: That's true, I'll reference the constant instead of repeating the value generation here.

tests/conftest.py
vshlapakov: Clear, makes sense to me, thanks for the explanation :ok_hand:

tests/conftest.py
victor-torres: Done.

tests/conftest.py
victor-torres: Done.

victor-torres:
@vshlapakov I've just updated the pull request.

Easy way to test the feature:

Rename some test case, for example, `tests/client/test_collections.py::test_truncate` to `tests/client/test_collections.py::test_truncate_new`.

Execute the test command specifying the environment variables (you can use any dummy Scrapy Cloud project):

```
HS_PROJECT_ID=<project id> HS_AUTH=<api key> AUTH=<api key> USER_AUTH=<api key> HS_ENDPOINT= DASH_ENDPOINT= py.test tests/client/test_collections.py::test_new_case
```

Check the new cassette file `tests/client/cassetes/test_collections/test_truncate_new.gz` with:
```
git status
```

vshlapakov:
That's just awesome, thank you so much :clap:

PR #116, opened 2019-04-19
Switch items API to iteration

piercefreeman:
The current `Items` iteration code requests all crawled data up-front.  For large enough crawls, this results in an indefinite hang from the API as it stalls to return all requests.


vshlapakov:
hey @piercefreeman thanks a lot, you helped to uncover a couple of old bugs - that's not the function implemented wrong, but a wrong usage of `_DownloadableProxyMixin` class that should have addressed it. I fixed it with #117, could you please check if master branch is fixed now before I release a patch?

vshlapakov:
Should be fixed in 2.1.1, thanks again!

PR #125, opened 2019-07-19
It adds the ability to cancel multiple jobs

rafaelcapucho:


scrapinghub/client/jobs.py
```diff
@@ -77,6 +79,64 @@ def count(self, spider=None, state=None, has_tag=None, lacks_tag=None,
...
+        """
+        update_kwargs(params, count=count, keys=keys)
+
+        if keys and count:
+            raise ValueError("keys and count can't be defined simultaneously")
+
+        elif not keys and not count:
+            raise ValueError("keys or count should be defined")
+
+        elif isinstance(keys, list) and len(keys) == 0:
```
ruairif: If keys can only be a `list` or `None` shouldn't this check be handled by `if not keys and not count`?

vshlapakov: +1, I think we can safely drop the condition, it's covered already

rafaelcapucho:
Hello guys @vshlapakov @ruairif ,
The PR is updated, with the suggested changes and the test cases, thx.

tests/client/test_job.py
```diff
@@ -44,6 +45,49 @@ def test_job_update_tags(spider):
     assert job2.metadata.get('tags') == ['tag2']
 
 
+def test_cancel_jobs(spider):
```
vshlapakov: Looks good, could you please split the large single test to a few simpler ones, and recreate cassettes? The small tests would help with getting a whole picture when something is broken.

rafaelcapucho: Good point, I'm going to change it :)

Issue #124, opened 2019-07-11
How to retrieve only items "_key". Is nodata supported in `items.iter()`?


manycoding:
I found an example with which I can retreive only the specified meta from job items:
https://doc.scrapinghub.com/api/items.html?highlight=metadata#examples

This could be handy to parallelize download and fix the issue with filters, collections (since we cannot properly divide items before download after `filter` is applied, or for collections which have disordered or duplicated keys) https://github.com/scrapinghub/arche/issues/13 by fetching only keys (I think it should be fast), breaking them on batches and then downloading the data.

Naturally, getting just keys should be much faster to solve the issues above. I tested locally with curl for 13kk keys:
CPU times: user 2.76 s, sys: 1.56 s, total: 4.33 s
Wall time: 9min 55s

That's not really fast, but anyway - is there `nodata`?

vshlapakov:
The following example works fine, does it answer your question?

```
In [11]: for item in job.items.iter(nodata=True, meta=['_key']):
    ...:     print(item)
    ...: 
{u'_key': u'123/45/67/0'}
```
I suppose we should update the docs for the client :thinking:

PR #129, opened 2019-08-19
Add nodata example to collections

manycoding:
For me it seems the most useful example. In particular, if we want to parallelize collections download we can first fetch all keys then split on batches.
I would also like to add `nodata` to both items and collections`iter()`, but I don't know how to do it.

### DIFF
--- a/scrapinghub/client/collections.py
+++ b/scrapinghub/client/collections.py
@@ -131,11 +131,10 @@ class Collection(object):
         ...     print(elem)
         [{'_key': '002d050ee3ff6192dcbecc4e4b4457d7', 'value': '1447221694537'}]
         
-    - get 1000th item key::
+    - get generator over item keys::
     
-        >>> import itertools
         >>> keys = foo_store.iter(nodata=True, meta=["_key"]))
-        >>> next(itertools.islice(keys, 1000, 1001))
+        >>> next(keys)
         {'_key': '002d050ee3ff6192dcbecc4e4b4457d7'}
 
     - filter by multiple keys, only values for keys that exist will be returned::
--- a/scrapinghub/client/collections.py
+++ b/scrapinghub/client/collections.py
@@ -133,8 +133,9 @@ class Collection(object):
         
     - get 1000th item key::
     
-        >>> keys = foo_store.list(nodata=True, meta=["_key"]))
-        >>> keys[1000]
+        >>> import itertools
+        >>> keys = foo_store.iter(nodata=True, meta=["_key"]))
+        >>> next(itertools.islice(keys, 1000, 1001))
         {'_key': '002d050ee3ff6192dcbecc4e4b4457d7'}
 
     - filter by multiple keys, only values for keys that exist will be returned::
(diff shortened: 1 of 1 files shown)


manycoding:
Closes #124

scrapinghub/client/collections.py
```diff
@@ -125,11 +125,17 @@ class Collection(object):
...
-    - iterate iterate over _key & value pair::
+    - iterate over _key & value pair::
 
         >>> for elem in foo_store.iter(count=1)):
         ...     print(elem)
         [{'_key': '002d050ee3ff6192dcbecc4e4b4457d7', 'value': '1447221694537'}]
+        
+    - get 1000th item key::
+    
+        >>> keys = foo_store.list(nodata=True, meta=["_key"]))
```
Gallaecio: For this, wouldn’t it be slightly better to use `.iter()` with https://docs.python.org/3/library/itertools.html#itertools.islice, to avoid building a big list?

scrapinghub/client/collections.py
```diff
@@ -125,11 +125,18 @@ class Collection(object):
...
 
         >>> for elem in foo_store.iter(count=1)):
         ...     print(elem)
         [{'_key': '002d050ee3ff6192dcbecc4e4b4457d7', 'value': '1447221694537'}]
+        
+    - get 1000th item key::
+    
+        >>> import itertools
+        >>> keys = foo_store.iter(nodata=True, meta=["_key"]))
+        >>> next(itertools.islice(keys, 1000, 1001))
```
vshlapakov: If your list has only 1k elements (and it's important because iter & list handlers return up to 1k elements max by default), this won't work:
```
>>> next(itertools.islice([1,2,3], 3, 4))
  File "<stdin>", line 1, in <module>
StopIteration
```
Also, getting Nth item's key is simpler (and more lightweight) by providing index param or getting item by the index, isn't it? In this case, you don't need to go through elements at all. There might be additional filters, so it probably won't work as you want in all cases, but as a general approach for an abstract scenario (i.e. as an example in the docs) that's a way better.
```
>>> job.items.list(index=1000)
{'field': 'data', ...}

>>> job.items.get(1000)
{'field': 'data', ...}
```
I'd suggest you to pick another example to demonstrate the nodata feature, smth like:
```
- iterate over keys only (doesn't fetch items data)

    >>> for elem in foo_store.iter(nodata=True, meta=["_key"])):
    ...     print(elem)
    {'_key': '002d050ee3ff6192dcbecc4e4b4457d7'}
```

manycoding: I thought it's obvious that we know our dataset size is bigger then 1000, isn't it?

> (and it's important because iter & list handlers return up to 1k elements max by default)
It seems they return all elements by default, and islice is the fastest way https://jupyter.scrapinghub.com/user/v/lab/tree/shared/Experiments/Arche/collections_nodata.ipynb

vshlapakov: > I thought it's obvious that we know our dataset size is bigger then 1000, isn't it?

Maybe, but  it's very common that users are glancing through the examples very fluently, and their assumptions can be totally different :slightly_smiling_face: Thanks!

vshlapakov:
Thanks @manycoding :+1:

Issue #121, opened 2019-05-04
msgpack errors when using iter() with intervals between each batch call


BurnzZ:
Good Day!

I've encountered this peculiar issue when trying to save up memory by processing the items in chunks. Here's a strip down version of the code for reproduction of the issue:

```
import pandas as pd

from scrapinghub import ScrapinghubClient

def read_job_items_by_chunk(jobkey, chunk=10000):
    """In order to prevent OOM issues, the jobs' data must be read in
    chunks.

    This will return a generator of pandas DataFrames.
    """

    client = ScrapinghubClient("APIKEY123")

    item_generator = client.get_job(jobkey).items.iter()

    while item_generator:
        yield pd.DataFrame(
            [next(item_generator) for _ in range(chunk)]
        )

for df_chunk in read_job_items_by_chunk('123/123/123'):
    # having a small chunk-size like 10000 won't have any problems

for df_chunk in read_job_items_by_chunk('123/123/123', chunk=25000):
    # having a bug chunk-size like 25000 will throw out errors like the one below
```

Here's the common error it throws:

```
<omitted stack trace above>

    [next(item_generator) for _ in range(chunk)]
  File "/usr/local/lib/python2.7/site-packages/scrapinghub/client/proxy.py", line 115, in iter
    _path, requests_params, **apiparams
  File "/usr/local/lib/python2.7/site-packages/scrapinghub/hubstorage/serialization.py", line 33, in mpdecode
    for obj in unpacker:
  File "msgpack/_unpacker.pyx", line 459, in msgpack._unpacker.Unpacker.__next__ (msgpack/_unpacker.cpp:459)
  File "msgpack/_unpacker.pyx", line 390, in msgpack._unpacker.Unpacker._unpack (msgpack/_unpacker.cpp:390)
  File "/usr/local/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x8b in position 67: invalid start byte
```

Moreover, it throws out a different error when using a much bigger chunk-size, like **50000**:

```
<omitted stack trace above>

    [next(item_generator) for _ in range(chunk)]
  File "/usr/local/lib/python2.7/site-packages/scrapinghub/client/proxy.py", line 115, in iter
    _path, requests_params, **apiparams
[...]

PR #133, opened 2019-10-04
implement new iter_by_chunks() in items

BurnzZ:
Provides convenient wrapper for #121


scrapinghub/client/items.py
```diff
@@ -59,3 +59,30 @@ def _modify_iter_params(self, params):
         if offset:
             params['start'] = '{}/{}'.format(self.key, offset)
         return params
+
+    def iter_by_chunks(self, chunksize=10000, *args, **kwargs):
```
BurnzZ: I have no idea what might be the best method name for this, so please suggest if you have something in mind.

scrapinghub/client/items.py
```diff
@@ -59,3 +59,30 @@ def _modify_iter_params(self, params):
...
+        generator of item chunks.
+
+        This is a convenient method for cases when processing a large amount of
+        items from a job isn't ideal in one go due to the large memory needed.
+        Instead, this allows you to process it chunk by chunk.
+
+        You can improve I/O overheads by increasing the chunk value but that
+        would also increase the memory consumption.
+
+        :return: an iterator over a list of elements.
```
manycoding:
```suggestion
        :return: an iterator over items, yielding lists of items.
```
If we use typing, it would be more obvious.

scrapinghub/client/items.py
```diff
@@ -59,3 +59,30 @@ def _modify_iter_params(self, params):
...
+        processed = 0
+        while True:
+            next_key = self.key + '/' + str(processed)
+            items = [
+                item for item in self.iter(
+                    count=chunksize, start=next_key, *args, **kwargs)
+            ]
+            yield items
+            processed += len(items)
+            if len(items) < chunksize:
```
manycoding: This one is tested by https://github.com/scrapinghub/python-scrapinghub/pull/133/files#diff-2ac22a76ea3536d23090269c351e0b76R50, right?

scrapinghub/client/items.py
manycoding: `job.items.list_iter(n=10_000)`
`job.items.iter_by(n=10_000)`
`job.items.iter_by_chunks(chunksize=10_000)`
I like the first two.

BurnzZ: I like the first one @manycoding as it's giving a better context of what's being returned overall, which is a generator of `list`s.

BurnzZ: I'll omit the **PEP 515** for now for backwards compatibility. It would be a great addition in 2020 nonetheless, when Python2 is deprecated.

BurnzZ:
Hi @vshlapakov, I was wondering if there was a guide on adding new the vcr cassettes for this repo? The missing cassettes for this new method is what's causing the test failures. Thanks!

scrapinghub/client/items.py
```diff
@@ -59,3 +73,30 @@ def _modify_iter_params(self, params):
         if offset:
             params['start'] = '{}/{}'.format(self.key, offset)
         return params
+
+    def list_iter(self, chunksize=10000, *args, **kwargs):
```
vshlapakov: I perfectly understand that the logic is a nice fit for iterating through huge items collections, but I'd suggest to have a lower default chunk size in the library, say 1k, for the general case. There're many parameters that should be taken into account when defining the specific value, like total amount of items and avg item size. Does it make sense?

scrapinghub/client/items.py
```diff
@@ -37,6 +37,20 @@ class Items(_DownloadableProxyMixin, _ItemsResourceProxy):
             'size': 100000,
         }]
 
+    - retrieve items via a generator of lists. This is most useful in cases
```
vshlapakov: Special thanks for the nice docstrings in the PR :heart:

scrapinghub/client/items.py
vshlapakov: I'd say the average computer has still 8GB nowadays, and many budget notebooks have only 4GB. And it's not only about this: this is a thin client, it should be lightweight and fast, I wouldn't expect it eating even a few GBs of memory by default without my explicit approval. Also, having a pretty large chunk size means you need a relatively good network, which isn't often the case, so the decision regarding the chunk size should be conscious.

My key point here is that the default settings should fit an average user' demands and expectations, not an experienced one processing millions items in batches. It seems we have different opinions on "an average user", I'm OK with that. But as an experienced user you can use whatever you need by setting a custom chunk size value, I don't see a problem here.

vshlapakov: @manycoding @BurnzZ Guys, to keep it clear: I don't enforce decreasing the default size by any cost, it's just something I'd like to get a bit more attention and discuss rather than accepting as-is. From a practical point of view, a chunk of 100k items by 10KB gives approximately 1GB of memory, is it something you'd expect as users of the client library using default settings? Maybe I'm missing something here?

manycoding: Seems like a good time to send a questionnaire.

As an average user, I would use `.iter()` or work in an environment with enough memory first.

BurnzZ: Hey gents! To chime in the discussion here, the default `chunksize` of `10_000` was unconsciously used since it was working for our particular project, which has pretty much a typical number of items that one would see in project. Size per item is pretty much average I would say.

Now when choosing the default `chunksize` value, I think we'll need to set it to whatever value that **upholds** the objective of this brand new method, which is to prevent OOM issues faced by users when processing large jobs.

Speaking of large jobs, I have worked on projects which has **1 MB per item** which was insane. Consuming the jobs on those types of projects would really benefit in this new method we're creating. This makes me lean into choosing a lower `chunksize` altogether:

- `chunksize` of **1_000** * **1 MB** per item = **~1 GB**
- `chunksize` of **10_000** * **1 MB** per item = **~10 GB**

Lastly, I'm not sure if physical computers or laptops should be the basis for our benchmarks in setting this value. Nonetheless, we'll need to set the bar low. With that in mind, I'm proposing that we use **Scrapy Cloud** as the basis, where **1 Scrapy Cloud** unit would have **1 GB of memory**. Typical max SC units at the moment would be **6** which gives us a maximum of **6GB of memory** per job. (_P.S. We're using this heavily in Scrapy Cloud jobs so it would really make sense for us to make this as the basis_)

In the example above, using `chunksize=10_000` would result in **OOM** issues as **10 GB > 6 GB SC Maximum**. With that, I'm leaning towards setting `chunksize=1_000` as the default.

But before doing that, what do you guys think? @vshlapakov @manycoding

BurnzZ:
Hi @vshlapakov, thanks for linking out #112. I've tried it out and works well on my local machine. However, when it's being run on CI it's going to fail. Mostly because when we empty out the env vars `HS_ENDPOINT` and `DASH_ENDPOINT`, it's going to use the live endpoints: https://github.com/scrapinghub/python-scrapinghub/blob/master/tests/conftest.py#L68

When running in CI, these aren't emptied out and thus will use the default endpoints like: https://github.com/scrapinghub/python-scrapinghub/blob/master/tests/conftest.py#L15.

I'm not sure how to proceed with this.

scrapinghub/client/items.py
```diff
@@ -59,3 +73,30 @@ def _modify_iter_params(self, params):
...
+        :return: an iterator over items, yielding lists of items.
+        :rtype: :class:`collections.Iterable`
+        """
+
+        processed = 0
+        while True:
+            next_key = self.key + '/' + str(processed)
+            items = [
+                item for item in self.iter(
+                    count=chunksize, start=next_key, *args, **kwargs)
```
manycoding: I tried the code and realised it doesn't support count and start. The first one is the most important I guess.

vshlapakov:
@BurnzZ Thanks for checking! I see what you mean, it seems it doesn't work reliably :confused: I'm in lack of time right now, will check it in detail next week and update in the thread. And Valeriy is right, we still need to support count/offset logic, please take a look.

scrapinghub/client/items.py
```diff
@@ -59,3 +73,30 @@ def _modify_iter_params(self, params):
...
+            ]
+            yield items
+            processed += len(items)
+            if len(items) < chunksize:
+                break
```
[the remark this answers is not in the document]
BurnzZ: I haven't considered this use case since we pretty much consume the entire items in a job. Much appreciated @manycoding! I've added some docstrings and test case as well.

Though the CI tests will fail because of the current project setup on `vcr-py`. @vshlapakov is it okay to mock this at the moment or should we standardize it to use cassettes like the current tests?

vshlapakov:
@BurnzZ Sorry that the issue took so long, I finally found some time to go quietly through the problem and have something to share. 

The case is pretty tricky. You correctly created the cassettes, and the normalizing process works fine. The real problem is that the normalization covers only requests data, not responses. So, it correctly forms the 1st run-spider request, VCR.py finds the related snapshot and provides the saved response, but the response has a real job key with a real project ID as a part of its key. As a result, the subsequent requests done using the related job instance have different paths and can't be matched with the saved snapshots which leads to the tests failure. 

I [made an attempt](https://github.com/scrapinghub/python-scrapinghub/commit/0b7747e82a4feb4583c0a2bd77f3f649605873e0) on top of your changes to fix the specific case by normalizing the job instances in tests, it works [fine](https://travis-ci.org/scrapinghub/python-scrapinghub), feel free to use it to finish your changes. I have some doubts if this is the best approach to solve the puzzle, but that's clearly a completely different topic which will be addressed later.

BurnzZ:
Thanks @vshlapakov! I `cherry-picked` your commit and updated my tests. Everything passes now! 🎉

tests/client/test_items.py
```diff
@@ -36,3 +39,45 @@ def test_items_list(spider, json_and_msgpack):
...
+        next(o)
+
+
+def test_items_list_iter_with_start_and_count(spider, json_and_msgpack):
+    job = spider.jobs.run(meta={'state': 'running'})
+    job = normalize_job_for_tests(job)
+    _add_test_items(job, size=10)
+    job.finish()
+
+    o = job.items.list_iter(chunksize=3, start=3, size=7)
```
manycoding: I noticed one line is not executed by tests https://codecov.io/gh/scrapinghub/python-scrapinghub/compare/92ff0968871c95dd00b078bcf0f6f5eb944c0f2c...c4ef768d3dda78a6de3fa119db0652ec891e84fa/diff

```suggestion
    o = job.items.list_iter(chunksize=3, start=3, count=7)
```

BurnzZ: Thanks for spotting @manycoding ! Much appreciated! just fixed it now and all CI tests are passing. 🎉

scrapinghub/client/items.py
```diff
@@ -37,6 +39,34 @@ class Items(_DownloadableProxyMixin, _ItemsResourceProxy):
...
+        >>> next(gen)
+        Traceback (most recent call last):
+          File "<stdin>", line 1, in <module>
+        StopIteration
+
+    - retrieving via meth::`list_iter` also supports the `start` and `count`.
+      params. This is useful when you want to only retrieve a subset of items in
+      a job. The example below belongs to a job with 10 items::
+
+        >>> gen = job.items.list_iter(chunksize=2, start=5, size=3)
```
vshlapakov:
```suggestion
        >>> gen = job.items.list_iter(chunksize=2, start=5, count=3)
```

scrapinghub/client/items.py
```diff
@@ -59,3 +89,41 @@ def _modify_iter_params(self, params):
...
+        count = kwargs.pop("count", sys.maxsize)
+        processed = 0
+
+        while True:
+            next_key = self.key + "/" + str(start)
+            items = [
+                item for item in self.iter(
+                    count=chunksize, start=next_key, *args, **kwargs)
+            ]
+            yield items
```
vshlapakov: Just as a side note, when the total amount of items >= `count` and `count` isn't divisible by `chunksize`, with the current logic the function will return amount of items >= `count`. Say, `list_iter(chunksize=2, count=3)` will return 2 batches by 2 items, so 4 items in total, not 3. This is probably fine, but if keeping it as-is, I'd ask to at least mention the behaviour in the function docstring.

scrapinghub/client/items.py
BurnzZ: Thanks for catching @vshlapakov! I think this was when I was deciding to rename `count` as `size`. 😅

scrapinghub/client/items.py
BurnzZ: Thanks @vshlapakov! Turns out there was a logic that @manycoding suggested before that wasn't added to this PR. This has been fixed and I've added a test case for it.

vshlapakov:
@BurnzZ Awesome, thanks, it's a pleasure to work with you! :rocket: 
@manycoding Thanks a lot for help in CR and your suggestions! :raised_hands:

BurnzZ:
Thanks for all the help @vshlapakov @manycoding! ❤️  Can't wait to use this in most of our projects! 🎉

PR #140, opened 2019-12-11
parse_job_key: Improve ValueError messages

Gallaecio:
Include the offending value in ValueError exceptions from parse_job_key().

I got this in a project, and knowing what the value contained may have helped:

```
  […]
  File "/app/src/scrapinghub/scrapinghub/client/__init__.py", line 99, in get_job
    project_id = parse_job_key(job_key).project_id
  File "/app/src/scrapinghub/scrapinghub/client/utils.py", line 59, in parse_job_key
    raise ValueError("Job key should be a string or a tuple")
ValueError: Job key should be a string or a tuple
```

### DIFF
--- a/scrapinghub/client/utils.py
+++ b/scrapinghub/client/utils.py
@@ -56,14 +56,17 @@ def parse_job_key(job_key):
     elif isinstance(job_key, six.string_types):
         parts = job_key.split('/')
     else:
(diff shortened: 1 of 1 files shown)


vshlapakov:
Thanks @Gallaecio :+1:

PR #159, opened 2021-10-20
Drop versions earlier than Python 3.7 and update requirements

apalala:
### library upgrades
This update nominally drops support for Python 2.7, 3.5, and 3.6, and tests support for 3.10 to avoid libraries being pinned to very old versions, many of them with bugs or with security issues.

It's "nominally" because the code hasn't been changed except for deprecations enforced in Python 3.9 or 3.10.

### disabled tests
Tests that required running test servers are disabled:
* Running the servers locally is too complicated
* There are no changes to the library's logic. Only required library versions were changed

The tests can be re-enabled by someone with access to test servers.


.envrc
```diff
@@ -0,0 +1 @@
+export SH_APIKEY=''
```
Gallaecio, line 1: Why is this file Git-tracked?

pytest.ini
```diff
@@ -1,2 +1,10 @@
...
-addopts = --doctest-glob='scrapinghub/*.py'
+addopts =
+    --doctest-glob='scrapinghub/*.py'
+    --ignore=tests/client/test_items.py
+    --ignore=tests/client/test_job.py
+    --ignore=tests/client/test_logs.py
+    --ignore=tests/client/test_requests.py
+    --ignore=tests/client/test_samples.py
+    --ignore=tests/client/test_spiders.py
+    --ignore=tests/hubstorage/test_retry.py
```
Gallaecio, line 10: Why are we ignoring test files?

If the issue is that a newer pylint includes new checks, I think it would be better to configure Pylint to ignore those checks until we get around to address them. But checks that test files already pass should continue to be checked for new code, so ignoring the whole files is, in my opinion, not the best approach.

requirements-test.txt
```diff
@@ -1,8 +1,7 @@
...
 mock
-vcrpy==1.10.3
-# FIXME remove the constraint after resolving
-# https://github.com/pytest-dev/pytest/issues/2966
-pytest<3.3.0
-pytest-cov<2.6.0
-pytest-catchlog
-responses==0.10.6
+vcrpy
+pytest
```
Gallaecio, line 4: Based on our experience with a few open source projects, it’s best to freeze pytest, so that CI will not suddenly start failing in `master` due to new checks being added. Instead, we can make it a habit to update it after a release, for example, either addressing or temporarily ignoring any new check added.

requirements.txt
```diff
@@ -1,2 +1,24 @@
...
+requests==2.26.0
+    # via scrapinghub (setup.py)
+retrying==1.3.3
+    # via scrapinghub (setup.py)
+six==1.16.0
+    # via
+    #   retrying
+    #   scrapinghub (setup.py)
+urllib3==1.26.6
+    # via requests
```
Gallaecio, line 24: What about removing this file entirely? As a library, we should keep dependencies in `setup.py` and let them be as flexible as possible.

requirements-test.txt
```diff
@@ -1,8 +1,7 @@
...
+pytest-cov
+responses
+tox
```
Gallaecio, line 7: It probably makes sense to remove this file and move most of its contents to `tox.ini` (minus `tox` itself, whose installation should be covered in the README).

.bumpversion.cfg
```diff
@@ -1,8 +1,7 @@
 [bumpversion]
-current_version = 2.3.1
+current_version = 3.0.0
```
Gallaecio, line 2: I think it makes sense to keep version bumping out of pull requests. The `version2bump` command already handles this change, and also tags the release, so to use `version2bump` the person handling the next release would actually need to undo this change locally before running the command, otherwise `versin2bump major` would set the version to 4.0.0 instead of 3.0.0.

scrapinghub/client/collections.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import collections
+import collections.abc
 
 from six import string_types
```
Gallaecio, line 5: Should we remove this as part of this pull request, or will we handle the post-Python-2 cleanup separately?

scrapinghub/client/collections.py
```diff
@@ -160,7 +161,7 @@ def get(self, key, **params):
         """Get item from collection by key.
 
         :param key: string item key.
-        :param \*\*params: (optional) additional query params for the request.
+        :param **params: (optional) additional query params for the request.
```
Gallaecio, line 164: Did you check how this renders in the documentation?

setup.py
```diff
@@ -26,20 +26,20 @@
     platforms=['Any'],
     packages=['scrapinghub', 'scrapinghub.client', 'scrapinghub.hubstorage'],
     package_data={'scrapinghub': ['VERSION']},
-    install_requires=['requests>=1.0', 'retrying>=1.3.3', 'six>=1.10.0'],
-    extras_require={'msgpack': mpack_required},
+    install_requires= mpack_required + [
```
Gallaecio, line 29: Why is msgpack no longer optional?

setup.py
```diff
@@ -12,8 +12,8 @@
 
 is_pypy = '__pypy__' in sys.builtin_module_names
 mpack_required = ['msgpack>=1.0.0']
-if is_pypy:
-    mpack_required.append('msgpack-pypy>=0.0.2')
+# if is_pypy:
+#    mpack_required.append('msgpack-pypy>=0.0.2')
```
Gallaecio, line 16: I see the project has been abandoned and there is no release since 2013, so I’m OK with removing this, but why just comment out?

setup.py
```diff
@@ -26,20 +26,20 @@
...
         'Development Status :: 5 - Production/Stable',
         'License :: OSI Approved :: BSD License',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
```
Gallaecio, line 41: 3.6 end of life: 23 Dec 2021. So I think we should either keep it or hold this change until that date.

tests/hubstorage/test_jobq.py
```diff
@@ -177,6 +177,7 @@ def test_summary_countstart(hsproject):
                 [o['key'] for o in s2['summary'][-6:-3]])
 
 
+@pytest.mark.skip(reason='Servers not running')
```
Gallaecio, line 180: Can you elaborate? What do we need to do to get these tests working?

jdemaeyer:
`shub` no longer works on Arch Linux as Python 3.10 landed in the official repos a week ago (breaking wihle trying to run the no-longer-possible imports from `collections` instead of `collections.abc`). There has been no activity on this PR for two months, any chance it will be merged soon?

pawelmhm:
we need to move this forward, tests are failing on storage.vm.scrapinghub. I'm not sure why we have to rely on making connections to some development server, this is a recipe for trouble, is there no way to fake tests here? 

I also don't think we need to bother about upgrading all libraries and dropping python 2 support NOW. We need to have this working with python 3.10 asap and worry about other things later. I'll try to make other smaller PR with only essential things

apalala:
@pawelmhm, the upgrade on this pull request is all we need. I know because I've tested it enough.

Everything else can wait.

setup.py
Gallaecio: I have just found out that removing the entry from `extras_require` is backward incompatible, so I suggest we keep it with an empty list of requirements, as suggested [in the setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies):

> Best practice: if a project ends up not needing any other packages to support a feature, it should keep an empty requirements list for that feature in its extras_require argument, so that packages depending on that feature don’t break (due to an invalid feature name).

Gallaecio: https://github.com/scrapinghub/python-scrapinghub/pull/164

(I would usually provide a suggestion, but I needed a fork I could clone for immediate usage)

elacuesta:
Python 3.10 support added in #166

Issue #160, opened 2021-10-24
Some imports from standard lib collections are breaking on python 3.10


lbmendes:
Hi everyone,

Based on an issue from another repo (https://github.com/okfn-brasil/querido-diario/issues/502), I noticed that scrapinghub is using some imports from standard lib collections that are deprecated and not working on Python 3.10.

In Python 3.8 I have these results on ipython console:
~~~python
In [1]: from collections import Iterator
<ipython-input-1-4fb967d2a9f8>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import Iterator

In [2]: from collections import Iterable
<ipython-input-2-c0513a1e6784>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import Iterable

In [3]: from collections import MutableMapping
<ipython-input-3-069a7babadbf>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import MutableMapping
~~~

According to this, it is necessary to change the imports of Iterable, Iterator and MutableMapping to get these items from "collections.abc" instead of just "collections"

Here are the list of imports that I found:
 - tests/client/test_job.py - from collections import Iterator
 - tests/client/test_frontiers.py - from collections import Iterable
 - tests/client/test_projects.py - from collections import defaultdict, Iterator
 - scrapinghub/hubstorage/resourcetype.py - from collections import MutableMapping

Gallaecio:
Fixed by https://github.com/scrapinghub/python-scrapinghub/pull/166 and released in 2.4.0.

PR #166, opened 2022-03-09
Python 3.10 compatibility

elacuesta:
Update some imports and test requirements for py310 compatibility.

Small overlap with #159, I think we can merge this sooner.


elacuesta:
For some reason, any version of `pytest` greater than `3.7.0` breaks the tests with
`requests.exceptions.ConnectionError: HTTPConnectionPool(host='storage.vm.scrapinghub.com', port=80): Max retries exceeded with url: /ids/2222222/spider/hs-test-spider?create=True (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f05c493fbe0>: Failed to establish a new connection: [Errno 113] No route to host'))`

That is a problem for py310, because of https://github.com/pytest-dev/pytest/pull/8540, which is available since [`6.2.4`](https://docs.pytest.org/en/stable/changelog.html#pytest-6-2-4-2021-05-04)

elacuesta:
> I see `E AttributeError: module 'collections' has no attribute 'Iterable'` in the latest test failures, which sounds like #160

That particular error should have been fixed by 1f6f39a47adbdc343fad24eae61e563c67a1be67, I don't see it in the most recent runs.

Gallaecio:
https://github.com/scrapinghub/python-scrapinghub/blob/c70137acf2bec5f86d1eac38605dd7ffe68d1e52/scrapinghub/client/collections.py#L188

elacuesta:
:facepalm: 
Thanks, should be fixed by a29a80f4af67eed450eb92f489f7036b15884c4f. The `requests.exceptions.ConnectionError` problem seems to be still present though :sweat:

Gallaecio:
Backporting the Python 3.10 fix I have been able to identify https://github.com/pytest-dev/pytest/pull/3754 as the cause of the issue. Things work with the prior upstream commit.

I have pushed at https://github.com/Gallaecio/pytest/tree/3.7.0-3.7.1-fork the latest working pytest with the Python 3.10 fix backported, in case you want to use that in the meantime.

I will continue looking into it for a more permanent fix.

requirements-test.txt
```diff
@@ -1,8 +1,6 @@
 mock
-vcrpy==1.10.3
-# FIXME remove the constraint after resolving
-# https://github.com/pytest-dev/pytest/issues/2966
-pytest<3.3.0
-pytest-cov<2.6.0
-pytest-catchlog
-responses==0.10.6
+vcrpy
+# FIXME vcrpy cassette fails after this version
```
Gallaecio, line 3:
```suggestion
```

:)

scrapinghub/client/collections.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
-import collections
 
 from six import string_types
+from six.moves import collections_abc
```
Gallaecio, line 4: It does not need to happen in this pull request, but it may be time to remove dependency on six altogether, since we are stopping Python 2 support.

Gallaecio (approved):

PR #177, opened 2024-12-13
Support the SHUB_APIURL and SHUB_STORAGE environment variables

Gallaecio:


scrapinghub/client/__init__.py
```diff
@@ -27,17 +27,23 @@ def request(self, *args, **kwargs):
...
-        provide a Scrapinghub APIKEY through this argument or deploying ``SH_APIKEY``.
-    :param dash_endpoint: (optional) Scrapinghub Dash panel url.
+        provide a Scrapy Cloud API key through this argument or deploying
+        ``SH_APIKEY``.
+    :param dash_endpoint: (optional) Scrapy Cloud API URL.
+        If not provided, it will be read from the ``SHUB_APIURL`` environment
+        variable, or fall back to ``"https://app.zyte.com/api/"``.
+    :param endpoint: (optional) Scrapy Cloud storage API URL.
+        If not provided, it will be read from the ``SHUB_STORAGE`` environment
+        variable, or fall back to ``"https://storage.scrapinghub.com/"``.
```
elacuesta, line 46: I'm not sure how to handle this but the `endpoint` parameter is not handled by the `ScrapinghubClient` directly, it's one of the additional arguments passed as `**kwargs` to the `HubstorageClient` class.

Gallaecio:
@elacuesta I am having a hard time figuring out why tests get a gzip-decoding issue on PyPy. I wonder if it is OK to merge stuff and release despite that or if we should look into it, e.g. maybe some dep upgrade is causing issues.

Gallaecio:
Oh, it’s vcrpy 6+ that broke that.

elacuesta (approved):

PR #178, opened 2024-12-13
Freeze vcrpy

Gallaecio:


Gallaecio:
I think the only failing test is a false positive due to limited resources on the CI infrastructure. `tox -e pypy310-msgpack -- tests/hubstorage/test_system.py::test_succeed_with_close_reason` works locally, although it does take 1m 40s to run :|.

PR #180, opened 2025-02-05
Fix coverage upload.

wRAR:


wRAR:
> AttributeError: 'VCRHTTPResponse' object has no attribute 'version_string'

Gallaecio:
> > AttributeError: 'VCRHTTPResponse' object has no attribute 'version_string'

Rings a bell. In any case, out of scope here.

